PR #10, opened 2024-03-22
feat: create Tips contracts

farrellh1:
Create contract that able to do:
- Push tip (send tip directly to wallet address)
- Send tip (send tip to the contract address, and the employee can withdraw later)
- Withdraw (withdraw tip as employee)
- GetTips (get tips earned by an address)
- Add events PushTip, SendTip, TipWithdrawal


contract/test/Tips.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      const member1StartingBalance = await ethers.provider.getBalance(member1.address);
+
+      await tips.connect(member1).withdraw();
+
+      const member1EndingBalance = await ethers.provider.getBalance(member1.address);
+
+      expect(member1EndingBalance).to.gt(member1StartingBalance);
+      expect(await tips.getTips(member1.address)).to.equal(0); // Balance should be reset
+    });
```
farrellh1, line 97: @hermannleboss , i wrote the expectation like this
`expect(member1EndingBalance).to.gt(member1StartingBalance);`

because i tried to write like this
```javascript
const amountPerAddress = ethers.parseEther("10");
const member1StartingBalance = await ethers.provider.getBalance(member1.address);
await tips.connect(member1).withdraw();

const member1EndingBalance = await ethers.provider.getBalance(member1.address);
const member1ExpectedBalance = member1StartingBalance + amountPerAddress;

expect(member1EndingBalance).to.equal(member1ExpectedBalance);
```
but it's failing, let me know if there is another solution, thanks

contract/contracts/Tips.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+contract Tips {
+    mapping(address => uint256) private addressToTips;
```
hermannleboss, line 5: `mapping(address => uint256) private addressToTips;` 
This mapping is storing a uint256 value for each address. So you should name it balance.

contract/contracts/Tips.sol
```diff
@@ -0,0 +1,62 @@
...
+    );
+    event SendTip(
+        address from,
+        address[] to,
+        uint256 totalAmount,
+        uint256 amountPerAddress
+    );
+    event TipWithdrawal(address to, uint256 amount);
+
+    function pushTip(address[] memory _toAddresses) positiveAmountRequired external payable {
```
hermannleboss, line 21: rename _toAddresses into teamsMembersAddress
Variable & function nane need to be clear so when someone read your code he can understant what is the purpose of your variable and function.

contract/contracts/Tips.sol
```diff
@@ -0,0 +1,62 @@
...
+        address[] to,
+        uint256 totalAmount,
+        uint256 amountPerAddress
+    );
+    event SendTip(
+        address from,
+        address[] to,
+        uint256 totalAmount,
+        uint256 amountPerAddress
+    );
```
hermannleboss, line 18: you can rename `to` into `teamMembers `to so it's clear

contract/contracts/Tips.sol
```diff
@@ -0,0 +1,62 @@
...
+
+        bool tipSuccess = payable(msg.sender).send(tipEarned);
+        require(tipSuccess, "Failed to withdraw tips.");
+
+        addressToTips[msg.sender] = 0;
+
+        emit TipWithdrawal(msg.sender, tipEarned);
+    }
+
+    function getTips(address _address) external view returns (uint256) {
```
hermannleboss, line 54: Rename into getBalance to make it more clear

contract/test/Tips.ts
```diff
@@ -0,0 +1,108 @@
...
+
+describe("Tips", function () {
+  let tips: Tips
+  let companyOwner: SignerWithAddress, member1: SignerWithAddress, member2: SignerWithAddress;
+  let recipientAddress: Array<string>;
+  const TIP_AMOUNT = ethers.parseEther("20");
+
+  beforeEach(async function () {
+    // Get signers for testing accounts
+    [companyOwner, member1, member2] = await ethers.getSigners();
```
hermannleboss, line 14: Why you have `companyOwner `it's not evident that only the company owner can send or push tips. 
You can rename it into `sender`

contract/test/Tips.ts
hermannleboss: When i run your test it's working.

hermannleboss (changes requested):
Update requests and lets merge it.

hermannleboss (approved):

PR #14, opened 2024-03-26
Feature/integrate siwe in front end

aliassan:
Hie

You can check out the implementation for SIWE integration for front-end here. It can still be further refactored I think but this will do for demonstration purposes for now.

We will iterate and improve in this.


app/src/services/userService.ts
hermannleboss, line 1: Good, but the nonce is not related to the user.

hermannleboss (commented):
For the moment, i like the idea of the architecture, but i want to push you more far.
- Inside your services your are fetching data from the API
- For authServices abstraction of auth methods, The basics idea is good, but you miss something. 


I think, you should have one interface like authServices then you call methods authWithSIWE, authWithECDS, and the  in the same interface you can get information like is authenticate and logout. 


So instead of extending the interface you creat auth methods inside. 
So with the auth object, you juste have one import.

My requests are this : 
Restructure authServices : I'm open to discussion.
Optional: Structure API request : This can be donne in another PR.

aliassan:
Okay. thanks for the review! However, SIWE, ECDS, Email, OAuth etc are implementations or concrete classes and AuthService is an interface or abstraction in the traditional sense of the word in OOP, I think there are some subtle differences when it comes to how it is used in TypeScript.

SOLID 101 is the abstractions shouldn't know anything about the concrete implementations, if those go in the Interface it would violate the following:

1. **Single Responsibility** - The interface would become a "fat class" that knows too much, everything in this case
2. **Open-Closed** - A modification to any of the implementations would inevitably mean a change in the Interface and all the actors or users that depend on it as well as the other implementations. The interface or architecture should be closed for modification.
3. **Interface Segregation** - You would be forcing users to depend on functionality they don't use or need, e.g. a user who wants Email would be forced to indirectly have dependencies on SIWE, OAuth, ECDS and so forth.

If there's functionality you feel is redundant like `isAuthenticated` and `logout`, the interface can become an abstract class and provide default implementations for those such that any of the sub-classes will inherit that default behaviour and not have to implement them if they choose not to.

You'll notice I didn't abstract the functionality of `LoginView` itself, that can be abstracted as well such that the functionality doesn't know about what implementation(s) it uses, in that case we could use Dependency Injection to provide the particular implementation(s) we want based on the user interface. 

If the implementations are included in the interface then that won't be possible or rather of no use. The auth object will and should indeed have one import in use cases, controllers, presenters and gateways but it will have relevant concrete imports in details such as the UI because the assumption is the UI knows what detail it requires.

aliassan:
For API I'm not sure I understand, are you suggesting to have a separate API service that handles the API calls or abstracts it so you can use whatever library you prefer to make the calls like Axios, native fetch etc. Maybe if you provide an example it will help.

app/src/services/userService.ts
aliassan: It was following the Ethereum structure of including the `nonce` in the account data. In that structure the latest valid nonce is stored persistently in the database. You can checkout this article here:

https://www.toptal.com/ethereum/one-click-login-flows-a-metamask-tutorial

hermannleboss:
> For API I'm not sure I understand, are you suggesting to have a separate API service that handles the API calls or abstracts it so you can use whatever library you prefer to make the calls like Axios, native fetch etc. Maybe if you provide an example it will help.

Here the idea is to have an abstration of API call, so you can whatever for the implementation. 

A structure like this

V1
src/
|-- api/
|   |-- UserAPI
|   |-- TeamAPI
|-- services/
|   |-- ApiService.js

The abstration of API service is for the fetch methods, you can use whatever library you prefer to make the calls like Axios, native fetch etc



src/
|-- api/
|   |-- UserAPI
|   |-- TeamAPI
|-- firebase/
|   |-- RandomEntity.js
|   |-- ...
|-- L8/
|   |-- RandomEntity.js
|   |-- ...
|-- services/
|   |-- ApiService.js
|   |-- DataService.js
|   |-- FirebaseService.js



The abstraction her is to abtract the source of data. But the implementation is not clear

hermannleboss:
> Okay. thanks for the review! However, SIWE, ECDS, Email, OAuth etc are implementations or concrete classes and AuthService is an interface or abstraction in the traditional sense of the word in OOP, I think there are some subtle differences when it comes to how it is used in TypeScript.

Look at this structure 


// Define an interface for AuthService
interface AuthServiceInterface {
  authenticateUser(): Promise<string>;
}
export class AuthService {
  token: string;
   constructor(token: string) {
    this.token = token;
  }

  isAuthenticated(): boolean {
    // Check if the authentication token is present in local storage
    return localStorage.getItem('authToken') !== null;
  }

  logout(): void {
    // Remove the authentication token from local storage
    localStorage.removeItem('authToken');
  }
}


// Implement SIWE authentication
export class SIWEAuthService implements AuthServiceInterface {
  async authenticateUser(): Promise<string> {
    // Implementation of SIWE authentication
    // Authentication logic using ECDSA signature
    return 'example_token'; // For demonstration purposes only
  }
}

// This is more simple export an utility function that authenticate the user

export function SIWEFunc() {
  // Auth with SIWE then return the function. 

  return 'example_token'; // For demonstration purposes only 
}

app/src/services/userService.ts
aliassan: Yes, we will still be using SIWE package, just slightly different than the example implementation. You had said nonce is not related to user but my suggestion is nonce is related to user the same way password is related to user in an email/password setup, it's part of user authentication. It's similar to the SIWE package example just we persist the nonce in a database instead of using a session library middleware which stores it in memory.

aliassan:
> > For API I'm not sure I understand, are you suggesting to have a separate API service that handles the API calls or abstracts it so you can use whatever library you prefer to make the calls like Axios, native fetch etc. Maybe if you provide an example it will help.
> 
> Here the idea is to have an abstration of API call, so you can whatever for the implementation.
> 
> A structure like this
> 
> V1 src/ |-- api/ | |-- UserAPI | |-- TeamAPI |-- services/ | |-- ApiService.js
> 
> The abstration of API service is for the fetch methods, you can use whatever library you prefer to make the calls like Axios, native fetch etc
> 
> src/ |-- api/ | |-- UserAPI | |-- TeamAPI |-- firebase/ | |-- RandomEntity.js | |-- ... |-- L8/ | |-- RandomEntity.js | |-- ... |-- services/ | |-- ApiService.js | |-- DataService.js | |-- FirebaseService.js
> 
> The abstraction her is to abtract the source of data. But the implementation is not clear

Okay, I think get what you mean, lemme try work something out for that.

aliassan:
> > Okay. thanks for the review! However, SIWE, ECDS, Email, OAuth etc are implementations or concrete classes and AuthService is an interface or abstraction in the traditional sense of the word in OOP, I think there are some subtle differences when it comes to how it is used in TypeScript.
> 
> Look at this structure
> 
> ```ts
> // Define an interface for AuthService
> interface AuthServiceInterface {
>   authenticateUser(): Promise<string>;
> }
> export class AuthService {
>   token: string;
>    constructor(token: string) {
>     this.token = token;
>   }
> 
>   isAuthenticated(): boolean {
>     // Check if the authentication token is present in local storage
>     return localStorage.getItem('authToken') !== null;
>   }
> 
>   logout(): void {
>     // Remove the authentication token from local storage
>     localStorage.removeItem('authToken');
>   }
> }
> 
> 
> // Implement SIWE authentication
> export class SIWEAuthService implements AuthServiceInterface {
>   async authenticateUser(): Promise<string> {
>     // Implementation of SIWE authentication
>     // Authentication logic using ECDSA signature
>     return 'example_token'; // For demonstration purposes only
>   }
> }
> 
> // This is more simple export an utility function that authenticate the user
> 
> export function SIWEFunc() {
>   // Auth with SIWE then return the function. 
> 
>   return 'example_token'; // For demonstration purposes only 
> }
> ```

Oh yes, you're right! My bad, I had extended the interface instead of implementing it, I'm still getting used to TypeScript. Yeah that's it, create an abstract class with default implementation that sub-classes will inherit. I've got some ideas about that, also with the views I want to have separate util functions to separate the logic from the views, I'll work on it.

hermannleboss:
@aliassan 
For the API structure and the SOLID implémentation, Plese juste make it work like it can be.
Somes architecture is difficult to implement on VueJS.

So juste make it work like you want. 

We can make refactoring when we realy know how to implement the right design princiles.

hermannleboss (changes requested):
Good enough,

- [x] Merge develop into your branch
- [ ] Fix merge conflict
- [ ] Run lint and format for Code Quality

Then will merge if the CI CD pass

app/src/adapters/web3LibraryAdapter.ts
```diff
@@ -0,0 +1,61 @@
...
+export class EthersJsAdapter implements IWeb3Library {
+  private provider: any
+  //private provider: ethers.providers.Web3Provider | null = null;
+  //private signer: ethers.Signer | null = null;
+  private signer: any
+
+  initialize(): void {
+    // Initialize provider
+    if ('ethereum' in window) this.provider = new BrowserProvider(window.ethereum as any)
+    //this.signer = this.provider.getSigner();
```
farrellh1, line 21: maybe u can fix like lint problem with this
```typescript
(window as any).ethereum
```

aliassan: Yeah, this passed the CI/CD test though. Initially it was failing because I was using `window.ethereum` direct but I changed to `window.ethereum as any` in `BrowserProvider` constructor which is similar to what you're suggesting.

Issue #12, opened 2024-03-23
feat: Implement Open API


hermannleboss:
Implementing OpenAPI in the backend of the project can bring several benefits:

1. **Standardization**: OpenAPI is a widely accepted standard for designing, building, and documenting APIs. By adopting this standard, we can ensure that our API design remains consistent and easy to understand.

2. **Documentation**: OpenAPI provides a clear and detailed documentation of the API endpoints, parameters, data models, etc. This can significantly reduce the learning curve for new developers and increase productivity.

3. **Client SDK Generation**: With OpenAPI specifications, we can automatically generate client SDKs for various languages. This can save a lot of development time and reduce the chances of errors.

4. **API Testing and Interactive Exploration**: Tools like Swagger UI can take an OpenAPI specification and generate interactive API documentation that lets developers test endpoints right from the browser.

5. **Integration with API Gateways**: Many modern API Gateways support importing OpenAPI specifications to quickly set up routes, models, and validation.

By implementing OpenAPI, we can make our backend more robust, scalable, and developer-friendly.

## Acceptance criteria

- [ ] Self generated API documentation using OpenAPI / Swagger

PR #18, opened 2024-03-29
feat: openapi-implementation

dasarathg68:
<h1> Description </h1>
The PR filed addresses issue #12 and presents OpenAPI specification for the Node.js backend.
swagger-autogen package is used to automatically parse the routes and identify paths, parameters and also descriptions.
The generated document is served using swagger-ui-express npm package on the /api-docs endpoint.
Further, the controller functions are grouped using by using tags. (Teams, Members)
Here's how it looks.
Screenshot 2024-03-30 at 12 07 56 PM
Screenshot 2024-03-30 at 12 08 31 PM

<h1> Type of Change </h1>
Non breaking change that implements OpenAPI for description of the APIs.


backend/package.json
```diff
@@ -3,6 +3,11 @@
   "version": "1.0.0",
   "description": "",
   "main": "src/index.ts",
+  "nodemonConfig": {
+    "ignore": [
+      "src/utils/swagger-output.json"
+    ]
+  },
```
hermannleboss, line 10: Why this is necessary

backend/src/controllers/memberController.ts
```diff
@@ -50,10 +56,13 @@ const updateMember = async (req: Request, res: Response) => {
...
 //delete members
 const deleteMembers = async (req: Request, res: Response) => {
+  /* 
+  #swagger.tags = ['Members']
+  */
   const { id } = req.params;
 
   try {
-    const member = await prisma.member.delete({
+    const member = await prisma.member.deleteMany({
```
hermannleboss, line 65: Why this change
And even if it's an optimization or a correction, it has nothing to do with swagger and open API.
So it shouldn't be in this PR

backend/package.json
hermannleboss, line 1: Look at this if it can be helpfull
https://www.npmjs.com/package/swagger-jsdoc

backend/src/utils/swagger-output.json
hermannleboss, line 1: If it's generated everytime we run the server do we need to have it inside git ?

hermannleboss (changes requested):
Good starting point. 

Look at this API,
https://petstore.swagger.io/


The api have to explain herself. 

- [x] Every input should have a type, 
- [x] we should know if it's optional or not. 
- [ ] And we should know the type of the output 
- [ ] and his status

Try to implement theses requirements

backend/package.json
dasarathg68: During dev we use nodemon to watch for changes right? 
It creates an infinite loop as every time the swagger-output is generated, it assumes it to be a change and restarts the server over and over again. 
Thats why this has to be present

backend/src/controllers/memberController.ts
dasarathg68: This change is about swagger.
swagger assumes anything in the pattern .delete() to be api endpoints.
So prisma.member.delete() would be considered to be an api endpoint which is not the case.

backend/package.json
dasarathg68: Thanks but I already took a look at that package and it cannot be used since we're not generating openapi specification. We're using swagger-autogen library which uses minimal directives to generate the specification automatically and swagger-ui-express to render the specification.

backend/src/utils/swagger-output.json
dasarathg68: We need the file to be present. It is a requirement of the swagger-autogen library.
If we don't add to git then we have to create it manually each time.
Atleast an empty file named swagger-output.json with just {} should be present

backend/package.json
hermannleboss: OK

dasarathg68:
> Good starting point.
> 
> Look at this API, https://petstore.swagger.io/
> 
> The api have to explain herself.
> 
> * [x]  Every input should have a type,
> * [x]  we should know if it's optional or not.
> * [x]  And we should know the type of the output
> * [x]  and his status
> 
> Try to implement theses requirements

@hermannleboss i have updated/reverted the code to work with the manual version and i have addressed the requirements. Please check the documentation at localhost:3000/api-docs after running the backend.

hermannleboss (approved):

PR #24, opened 2024-03-30
feat: add path for app folder

hermannleboss:
Path filtering for the github action


hermannleboss:
First commit with no update on app/** path. No pipeline executed

PR #28, opened 2024-04-11
Feature/integrate siwe in back end

aliassan:


backend/src/controllers/userController.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  try {
+    if (address) {
+      await prisma.user.create({
+        data: {
+          address,
+          nonce,
+        },
+      });
+
```
hermannleboss, line 26: You are not checking if the user existe or not. 
And you are not cheking if who whant to create a user for this address is the owner of the wallet.

backend/src/controllers/userController.ts
```diff
@@ -0,0 +1,82 @@
...
+      },
+    });
+
+    await prisma.$disconnect();
+
+    if (!user)
+      return res.status(200).json({
+        success: true,
+        nonce: false,
+      });
```
hermannleboss, line 68: I think you can generate a nonce for a user that not exit.

backend/src/script.ts
hermannleboss, line 1: I don't know why you have this file

hermannleboss (approved):

hermannleboss (changes requested):
Fix the merge conflict.
I juste have a doubt about the SIWE process let discuss tomorrow

backend/src/controllers/userController.ts
aliassan: Okay, noted about that

backend/src/controllers/userController.ts
aliassan: In that case there's no need to make an API request, it can be done in the front-end using the library. If you want users that don't exist to be able to interact with the application then that's okay but then anyone with the `siwe` library can generate valid data and there's no way of ensuring that it's valid for _our_ platform.

backend/src/script.ts
aliassan: I forgot to remove that but will do

hermannleboss: OK

backend/src/controllers/userController.ts
hermannleboss: No, I'm just thinking how a user who does not have an account can receive a nonce to sign his message.

hermannleboss (approved):

PR #39, opened 2024-04-17
Feature/implement siwe frontend and backend

aliassan:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

# Checklist:


aliassan:
@dasarathg68 I've created authentication functionality for you to test on the front-end and back-end routes. For back-end you can use `authorizeUser` middleware in `authMiddleware.ts` and for front-end you can use `isAuthenticated` in `AuthService` class from `authService.ts`. 

You can also call, `logout` or `getToken`. They are all static methods so you can call them directly on the class. You'll also need to set FRONTEND_URL and SECRET_KEY (any HS256 compatible key) environment variables in .env and BACKEND_URL constant in `utils.ts` in the front-end.

app/src/utils/util.ts
hermannleboss, line 1: It's better to use an env Variable for this.

app/src/apis/authApi.ts
```diff
@@ -1,22 +1,52 @@
+import { BACKEND_URL } from '@/utils/util'
+
```
hermannleboss, line 2: Refactor using an env Variable for this.

app/src/utils/util.ts
dasarathg68: @aliassan Farrell has already defined a folder for constants under /app/src/constant.
Maybe you can import the url from the env and export it as a variable as you are already doing.

dasarathg68:
You can specify the env variables used in /app/.env.example

backend/src/config/serverConfig.ts
```diff
@@ -35,7 +35,7 @@ class Server {
 
   private middleware() {
     this.app.use(express.json());
-    this.app.use(cors());
+    this.app.use(cors({origin: (process.env.FRONTEND_URL as string)}));
```
hermannleboss, line 38:
```suggestion
    this.app.use(cors({origin: (process.env.FRONTEND_URL ? process.env.FRONTEND_URL : "*")}));
```

in this case it's working if the FRONTEND_URL is not set.

What do you think. 
And a warning console log to say that you should set the FRONTEND_URL in the .env

hermannleboss (approved):

backend/src/config/serverConfig.ts
aliassan: I think allowing all origins would pause a security risk, especially when we're using JWT with localStorage, I would rather we force that someone specifies an origin, that's why we're using the `cors` middleware after all

hermannleboss: OK in this case log a warning to say you have to setup the FRONTEND_URL or just throw an error.

aliassan: I'm not sure I know how to implement that. Where would the warning be logged? If it's a problem we can just implement your simple fix.

hermannleboss: I will show you. 
Let create a fix for it later

hermannleboss (approved):

PR #43, opened 2024-04-17
feat: setup routes protection 

### LINKED ISSUE #25: feat: protect routes
Use auth to protect route in the frontend and bakend

### Acceptance criteria 

- [ ] Structure the database to have relation between user and team
- [ ] Secure Team CRUD Backend
- [ ] Secure Team CRUD Frontend

dasarathg68:
# Description

This PR addresses #25 and #44 and sets up route protection for frontend and backend using JWT.
Also redirects the user after SIWE

Fixes #25 and #44  

## Type of change

# Checklist:


app/src/stores/address.ts
```diff
@@ -0,0 +1,21 @@
...
+    ownerAddress: localStorage.getItem('ownerAddress') || '' // Load owner address from localStorage
+  }),
+  actions: {
+    setOwnerAddress(address: string) {
+      this.ownerAddress = address
+      localStorage.setItem('ownerAddress', address) // Save owner address to localStorage
+    },
+    getOwnerAddress() {
+      return this.ownerAddress
+    },
```
farrellh1, line 15: we don't have to create getters for it because we already create `ownerAddress` in state,

dasarathg68: ah yes. got it! thanks for pointing that out

app/src/components/NavBar.vue
aliassan, line 1: In `LoginView.ts` I introduced a convention where I separate the UI from the logic. I placed the logic in `utils` so in your case it would be `navBarUtil.ts` where you define `logout` and all the relevant dependencies. What would stay in `NavBar.ts` is the UI specific dependencies like `defineEmits`. You would then simply import `logout` from `navBarUtil.ts` and that's all. Your code in the component would look the same except for one extra import statement

dasarathg68: Thanks for that, I have resolved it!

app/src/stores/address.ts
```diff
@@ -0,0 +1,18 @@
...
+    setOwnerAddress(address: string) {
+      this.ownerAddress = address
+      localStorage.setItem('ownerAddress', address) // Save owner address to localStorage
+    },
+    clearOwnerAddress() {
+      this.ownerAddress = ''
+      localStorage.removeItem('ownerAddress') // Remove owner address from localStorage
+    }
+  }
+})
```
hermannleboss, line 18: Keep it and make a fix later

hermannleboss (approved):

PR #56, opened 2024-04-22
fix: redirect user after token expiry

### LINKED ISSUE #55: issue: No redirection when the token is not valid again
When the token validation date has passed, normaly it should redirect to login page, but don't.
image

You can't navigate.

dasarathg68:
# Description

Clears the token, logs the user out from the application and redirects it to /login by calling the logout method.

Fixes #55 

## Type of change

# Checklist:

### DIFF
--- a/app/src/apis/authApi.ts
+++ b/app/src/apis/authApi.ts
@@ -1,5 +1,5 @@
 import { BACKEND_URL } from '@/constant/index'
-
+import { logout } from '@/utils/navBarUtil'
 export interface IAuthAPI {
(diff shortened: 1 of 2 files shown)


app/src/apis/authApi.ts
```diff
@@ -19,6 +19,7 @@ export class AuthAPI {
       return true
     } else {
       //return false
```
farrellh1, line 21: i think the name of the function is typo, it should be `verifyToken` right?

dasarathg68: Yes, let me change that. 
Any other suggestions?

farrellh1 (approved):
LGTM 🚀

Issue #40, opened 2024-04-17
refactor: Refactoring the code to improve readability and upgradability


hermannleboss:
- Refactor svg icones into comopnents inside icone folder.
- Every component inside components should only display content and return event on interactions.
  - No API Call
  - To be determined

PR #58, opened 2024-04-22
refactor-codebase

### LINKED ISSUE #46: Issue: Validation to member address
Add validation to member address in the frontend and the backend.

dasarathg68:
# Description
Adds address validation to both frontend and backend and performs refactoring of the codebase

Fixes #46 
#40 

## Type of change

# Checklist:


farrellh1 (approved):
LGTM, wait for herman approval

app/src/utils/walletValidatorUtil.ts
hermannleboss, line 1: No need to create it

hermannleboss (changes requested):
Use ether.utils and emit event instead of using it in the component

dasarathg68 (commented):
Done with ethers.utils and refactoring

app/src/components/MemberCard.vue
```diff
@@ -1,80 +1,66 @@
...
+    <th>{{ member.id }}</th>
+    <th>{{ member.name }}</th>
+    <th>{{ member.walletAddress }}</th>
     <th>Action</th>
   </tr>
 
   <dialog
     id="my_modal_20"
+    v-if="showUpdateMemberModal"
     class="modal modal-bottom sm:modal-middle"
```
farrellh1, line 12: we can change this modal dialog to be component

app/src/components/MemberCard.vue
```diff
@@ -1,80 +1,66 @@
...
         |
-        <input type="text" class="grow" v-model="walletInput" />
+        <input type="text" class="grow" v-model="updateMemberInput.walletAddress" />
       </label>
       <div class="flex mt-2 justify-between">
-        <button class="btn btn-error size-sm" @click="deleteMember">Delete</button>
-        <button class="btn btn-primary" @click="updateMember">Update</button>
+        <button class="btn btn-error size-sm" @click="emits('deleteMember', updateMemberInput.id)">
+          Delete
+        </button>
```
farrellh1, line 35: and also we can refactor this button into single component, and the props will be the color of the button and the size and the function of the click

app/src/types/types.ts
```diff
@@ -11,3 +11,16 @@ export interface Member {
   walletAddress: string
   teamId: number
 }
+
+export interface MemberInput {
+  name: string
+  walletAddress: string
+  id?: string
+  isValid: boolean
+}
```
farrellh1, line 20: i think we can seperate each type into single file, just my opinion

farrellh1: and then add into `index.ts` to do barrel export https://medium.com/@klauskpm/do-a-barrel-export-aa5b79b76b05

farrellh1 (commented):
You should also remove the `console.log` because we shouldnt show any log in production

app/src/components/MemberCard.vue
dasarathg68: I think daisy ui already simplifies this for us and the whole point of Tailwind and Daisy UI is to define class names on the go.
A generic component, I feel, would take away the point of it.

backend/src/controllers/memberController.ts
```diff
@@ -14,21 +14,27 @@ const addMembers = async (req: Request, res: Response) => {
...
     const createdMembers = await prisma.member.createMany({
       data: membersToCreate,
     });
 
     res.status(201).json(createdMembers);
-  } catch (error) {
+  } catch (error: any) {
     console.error("Error adding members:", error);
-    res.status(500).json({ error: "Internal server error" });
+    res.status(500).json({ error: `${error.message}` });
```
aliassan, line 38: `errorResponse` util takes care of this for you including conditionally printing in dev mode. We standardised to avoid leaking sensitive information about the back-end architecture to the front-end

app/src/apis/memberApi.ts
```diff
@@ -56,9 +60,11 @@ export class FetchMemberAPI implements MemberAPI {
...
         throw new Error(`HTTP error! Status: ${response.status}`)
       }
-    } catch (error) {
+      show(ToastType.Success, 'Updated Member Details')
+    } catch (error: any) {
+      show(ToastType.Warning, error.message)
       console.error('Error:', error)
-      return
+      throw error
     }
```
aliassan: I feel you're violating Single Responsibility here, you're making API calls, displaying errors to the UI and handling errors at the same. You should just throw an error and have some lower level code somewhere decide what to do with the error. The way you're doing it forces you to go and implement this logic in all your API code instead of just where the API classes are used.

dasarathg68: The show function is from useToastStore (not a Vue function) and it keeps UI interactions separate from component logic, and although it might seem like error handling occurs within the component, show is still a lower-level code like you mentioned, fetched from the centralized Pinia store.

aliassan: Yes, you have a `try..catch` block around your code which is error handling. Instead you would have that where you are "using" `FetchMemberAPI` which is most likely going to be a composable soon although, right now you're calling it directly in a view.

app/src/apis/memberApi.ts
```diff
@@ -24,19 +24,22 @@ export class FetchMemberAPI implements MemberAPI {
...
     const { show } = useToastStore()
 
-    newMembers.map((member: Partial<Member>) => {
+    for (const member of newMembers) {
       if (!isAddress(member.walletAddress)) {
         useErrorHandler().handleError(new Error(`Invalid wallet address`))
```
aliassan, line 27: I suggest that you just throw the error and handle them where you're using or calling the classes, it's simpler.

hermannleboss (approved):

PR #75, opened 2024-05-07
refactor: tips and toast into composable

farrellh1:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/app/src/composables/tips.ts
+++ b/app/src/composables/tips.ts
@@ -1,0 +1,68 @@
+import { TIPS_ADDRESS } from '@/constant'
+import { ethers } from 'ethers'
+import ABI from '../abi/tips.json'
+import { onMounted, ref } from 'vue'
+import { useToast } from './toast'
+import { ToastType } from '@/types'
+import type { AddressLike } from 'ethers'
+
+export function useTips() {
+  const { showToast } = useToast()
+  const provider = new ethers.BrowserProvider((window as any).ethereum)
+
+  const contract = new ethers.Contract(TIPS_ADDRESS, ABI, provider)
+  const pushTipLoading = ref(false)
+  const sendTipLoading = ref(false)
+
+  async function pushTip(addresses: AddressLike[], amount: number) {
+    pushTipLoading.value = true
+
+    try {
+      const tx = await contract.pushTip(addresses, {
+        value: ethers.parseEther(amount.toString())
+      })
+      await tx.wait()
+
+      showToast(ToastType.Success, 'Tip pushed successfully')
+    } catch (error: any) {
+      showToast(ToastType.Error, error.reason ? error.reason : 'Failed to push tip')
+    } finally {
+      pushTipLoading.value = false
+    }
+  }
+
+  async function sendTip(addresses: AddressLike[], amount: number) {
+    sendTipLoading.value = true
+
+    try {
+      const tx = await contract.sendTip(addresses, {
+        value: ethers.parseEther(amount.toString())
+      })
+      await tx.wait()
+
+      showToast(ToastType.Success, 'Tip sent successfully')
+    } catch (error: any) {
+      showToast(ToastType.Error, error.reason ? error.reason : 'Failed to send tip')
+    } finally {
+      sendTipLoading.value = false
+    }
+  }
+
+  async function getBalance(address: AddressLike) {
+    return await contract.getBalance(address)
+  }
+
+  onMounted(async () => {
+    if (!(window as any).ethereum) showToast(ToastType.Info, 'Please install Metamask')
+
+    await provider.send('eth_requestAccounts', [])
+  })
+
+  return {
+    pushTipLoading,
+    pushTip,
+    sendTipLoading,
+    sendTip,
+    getBalance
+  }
+}
--- a/app/src/composables/toast.ts
+++ b/app/src/composables/toast.ts
@@ -1,0 +1,20 @@
+import { ToastType } from '@/types'
+import { ref } from 'vue'
+
+export function useToast() {
+  const type = ref(ToastType.Success)
+  const show = ref(false)
+  const message = ref('')
+
+  async function showToast(toastType: ToastType, toastMessage: string) {
+    type.value = toastType
+    show.value = true
+    message.value = toastMessage
+
+    setTimeout(() => {
+      show.value = false
+    }, 3000)
+  }
+
+  return { showToast }
+}


hermannleboss:
@farrellh1 One PR one Action. 

Things need to be separated and validate separately. 

Toast won't be a composable. Only Tips.

app/src/composables/tips.ts
```diff
@@ -0,0 +1,68 @@
...
+  async function pushTip(addresses: AddressLike[], amount: number) {
+    pushTipLoading.value = true
+
+    try {
+      const tx = await contract.pushTip(addresses, {
+        value: ethers.parseEther(amount.toString())
+      })
+      await tx.wait()
+
+      showToast(ToastType.Success, 'Tip pushed successfully')
```
hermannleboss, line 26: Separation of concern

Your composable Tips don't need to use Toast. 

For example if in the future you don't need to use toast every times you push tip but only when you push tips on a specific page.
How you will handle this ?
So you need to avoid to depend on it. But rather you can simply return the result and the component will decide what to do with this information.

app/src/composables/tips.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    try {
+      const tx = await contract.sendTip(addresses, {
+        value: ethers.parseEther(amount.toString())
+      })
+      await tx.wait()
+
+      showToast(ToastType.Success, 'Tip sent successfully')
+    } catch (error: any) {
+      showToast(ToastType.Error, error.reason ? error.reason : 'Failed to send tip')
```
hermannleboss, line 45: You should return the transaction and the error.
And the component will decide what to do in this case.

app/src/composables/tips.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  async function getBalance(address: AddressLike) {
+    return await contract.getBalance(address)
+  }
+
+  onMounted(async () => {
+    if (!(window as any).ethereum) showToast(ToastType.Info, 'Please install Metamask')
+
+    await provider.send('eth_requestAccounts', [])
+  })
```
hermannleboss, line 59: I think should be check on another side. 

SO you can simply remove it.

app/src/composables/tips.ts
```diff
@@ -0,0 +1,68 @@
...
+    } catch (error: any) {
+      showToast(ToastType.Error, error.reason ? error.reason : 'Failed to send tip')
+    } finally {
+      sendTipLoading.value = false
+    }
+  }
+
+  async function getBalance(address: AddressLike) {
+    return await contract.getBalance(address)
+  }
```
hermannleboss, line 53: Balance also need a loading state and should return error

app/src/composables/toast.ts
hermannleboss, line 1: You should remove this. 

Focus on Tips composable

hermannleboss (changes requested):

app/src/composables/tips.ts
aliassan: > Separation of concern
> 
> Your composable Tips don't need to use Toast.
> 
> For example if in the future you don't need to use toast every times you push tip but only when you push tips on a specific page. How you will handle this ? So you need to avoid to depend on it. But rather you can simply return the result and the component will decide what to do with this information.

Thank you for pointing this out, I've seen it used inside of API calls and a lot of places. If this functionality changes a lot of code will need to be updated/changed without any actual productivity, just maintenance

app/src/composables/tips.ts
aliassan: The check belongs in a utility class that does wallet management and should be triggered every time that a user tries to interact with MetaMask

aliassan (changes requested):
Can the tipping functionality be defined as a service in the services folder. We use it to track the application features and for separation of concerns. For instance authentication is a service or use case, tipping is also a service or use case.

When you look in the folder you immediately know the application features, if there happens to be problem, a change request or anything it's easier to implement and manage.

The composable then acts as a kind of controller interfacing between the UI and the `tips`service, it retrieves the functions from the service, creates the required reactive state and passes them on to the UI.

farrellh1:
so far i refactor it like this. but somehow the loading on push or send tip not reactive and maybe balanceLoading also not working. and i still have no idea how to implement this https://github.com/globe-and-citizen/cnc-portal/pull/75#discussion_r1592019020

app/src/composables/tips.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  return { balance, getBalance, loading, error }
+}
+
+export function usePushTip() {
+  const loading = ref(false)
+  const error = ref<any>(null)
+  const isSuccess = ref(false)
+
+  async function pushTip(addresses: string[], amount: number): Promise<void> {
```
hermannleboss, line 34: Return the transaction
Instead of returning the function directly, you return a function called execute that execute pushTip function and you can use the return value of pushTip function for the reactive value

app/src/composables/tips.ts
```diff
@@ -0,0 +1,119 @@
...
+      await tipsService.sendTip(addresses, amount)
+      isSuccess.value = true
+    } catch (err) {
+      error.value = err
+    } finally {
+      loading.value = false
+    }
+  }
+
+  return { sendTip, loading, error, isSuccess }
```
hermannleboss, line 66:
```suggestion
  const transaction= ref<any>(null) // Use the right type of transaction if exist
  const error = ref<any>(null)
  const isLoading = ref(false)
  const isSuccess = ref(false)

  async function sendTip(addresses: string[], amount: number): Promise<void> {
    try {
      loading.value = true
      await tipsService.sendTip(addresses, amount)
      isSuccess.value = true
    } catch (err) {
      error.value = err
    } finally {
      loading.value = false
    }
  }
  
   function execute(){
       transaction.value= await sendTip()
   }
  return { execute, transaction, IsLoading, error, isSuccess }
```

This is an example

hermannleboss: I use isLoading because you use IsSuccess, but for isSuccess i prefer isLoaded

app/src/main.ts
```diff
@@ -2,13 +2,22 @@ import './assets/main.css'
...
 
 app.use(createPinia())
+const toastOptions: PluginOptions = {
+  position: POSITION.BOTTOM_RIGHT,
+  pauseOnFocusLoss: false,
+  pauseOnHover: false,
+  timeout: 3000
+}
+app.use(Toast, toastOptions)
 app.use(router)
```
hermannleboss, line 21: To be revert

app/src/services/tipsService.ts
hermannleboss, line 1: LGTM

hermannleboss (changes requested):

hermannleboss:
> @farrellh1 One PR one Action.
> 
> Things need to be separated and validate separately.
> 
> Toast won't be a composable. Only Tips.

@farrellh1 i think you don't understand this comment. 

For every PR you need to do one and only on thing. It's simple for validation and easy for you to fix things. 

Right now you refactor Toast And Tips in the Same PR, Now Tips Refactor is good, and Toast one is not good.

I can't merge you PR

app/src/components/LoadingButton.vue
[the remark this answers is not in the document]
farrellh1: i want this component to have dynamic width or sizing, so we can define sizing only when we call the component

app/package.json
```diff
@@ -23,7 +23,8 @@
     "pinia": "^2.1.7",
     "siwe": "^2.1.4",
     "vue": "^3.4.15",
-    "vue-router": "^4.2.5"
+    "vue-router": "^4.2.5",
+    "vue-toastification": "^2.0.0-rc.5"
   },
```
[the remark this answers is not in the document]
farrellh1: yeah but i think we need a package to make easier to create stackable toast like u mentioned in last team meeting

app/src/composables/tips.ts
farrellh1: i still dont understand why we need to create seperate function between execute and sendTip. why dont we create execute that will do exactly what pushTip do?

hermannleboss: If you send tip in your component the result of send tip is not reactive. 

let say you have

```ts
 {sendTip, isloading}= useSendTips;
// then you will call the function send tips
const tx=await  sendTip()

Now you can't use the transaction as a reactiv state (because you have this value only when send tip is called)
```

The solution you have is to do 

```ts
const tx= ref(null)

tx.value=await sendTip()
```

As you can see it's more complicated

farrellh1: but we can do this instead
```typescript
async function sendTip(addresses: string[], amount: number): Promise<void> {
    try {
      loading.value = true
      tx.value = await tipsService.sendTip(addresses, amount)
      isSuccess.value = true
    } catch (err) {
      error.value = err
    } finally {
      loading.value = false
    }
  }
  ```

Comment (farrellh1) [reply app/src/composables/tips.ts:66]: and for the isLoaded i think it's better to use it on `getEvents` for the `isSuccess` for sending transaction

Comment (hermannleboss) [reply app/src/composables/tips.ts]: I don't understand

File: app/src/views/SingleTeamView.vue [hunk 9]
```diff
@@ -47,14 +47,7 @@
...
     </div>
-    <TipsAction
-      :addresses="team.members.map((member) => member.walletAddress)"
-      :pushTipLoading="pushTipLoading"
-      :sendTipLoading="sendTipLoading"
-      :tipAmount="tipAmount"
-      @pushTip="(addresses, amount) => pushTip(addresses, amount)"
-      @sendTip="(addresses, amount) => sendTip(addresses, amount)"
-    />
+    <TipsAction :addresses="team.members.map((member) => member.walletAddress)" />
```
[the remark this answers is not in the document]
farrellh1: yes, i done it because the loading will not working if we define the function in the parent and get the loading in children component. but i think it's possible to do it if we pass the loading from parent as a props, but in my opinion it makes in the component will be complex (IMO).

we dont have to use "emit events" since we already seperate the logic into the composables

app/src/composables/tips.ts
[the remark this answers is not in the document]
farrellh1: result of the transaction?

app/src/composables/tips.ts
```diff
@@ -0,0 +1,123 @@
...
+import type { EventResult, TipsEventType } from '@/types'
+import dayjs from 'dayjs'
+import type { Log } from 'ethers'
+import type { EventLog } from 'ethers'
+import { ref } from 'vue'
+
+const tipsService = new TipsService(EthersJsAdapter.getInstance())
+
+export function useTipsBalance() {
+  const balance = ref<string>('0')
```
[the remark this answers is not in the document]
farrellh1: i still dont understand this part, if we set it to null. then in `app/src/components/NavBar.vue` we need to do something like 
`balance ? balance.slice(0, 6) : 0`
which is same as setting it to 0 on ref declaration

```
For ex: for pushTip the transaction is by default null.
So if you check if the transaction is null that mean you don't have any transaction.

But if the balance is by default 0 and you receive the balance who is also 0, there is any difference between default one and 0 balance
```

hermannleboss (approved):

PR #85, opened 2024-05-09
feat/add network compatibility check

dasarathg68:
# Description

If the user is on a different network than the network specified in env.VITE_CURRENT_NETWORK_NAME while using the app, warn and ask the user to change network.

## Type of change

# Checklist:


farrellh1:
i just installed new library in this PR, maybe u could use it #75 or change base branch into my branch.

and also u need to update the docs for .env and .env.example

PR #95, opened 2024-05-10
Feat/set network for user

aliassan:
# Description

Adds a MetaMask utility class with method to switch network or add custom network for the user to the network required for the app.

Fixes # (issue)

## Type of change

# Checklist:


app/.env.example
```diff
@@ -1,3 +1,9 @@
 VITE_APP_TIPS_ADDRESS=
 VITE_APP_BACKEND_URL=
 VITE_APP_ETHERSCAN_URL=
```
farrellh1, line 3:
```suggestion
```

we can remove and use `VITE_APP_BLOCK_EXPLORER` instead, and dont forget to change the other code that use `VITE_APP_ETHERSCAN_URL`

hermannleboss:
I was thining that you can use a builder pattern for it. 
Something with default value and somes wiht custom value. 
Like with metamask, you have default network you can setup by chosing one and you can also configure a custom network.

app/.env.example
aliassan: Okay, if you have code that depends on that variable you can update once changes have been approved so I don't have to dig into the code looking for where it's used.

app/.env.example
```diff
@@ -1,3 +1,9 @@
...
+VITE_APP_BLOCK_EXPLORER=
```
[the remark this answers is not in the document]
aliassan: Yes I had requested that whoever is using that old value update their code once this is approved because I don't know which code is using that

app/src/constant/network.ts
hermannleboss, line 1: Your function are not clear here. 

First with your algo getNetwork return always sepolia

hermannleboss: And your code is not following the logic of what is need. 

- [ ] By default use Sepolia Network
- [ ] Is alias is set Use the alias Network
- [ ] If a custom network is set Use his configuration. 

But what you are doing

- Check the alias 
- if the alias don't exit Create the custom netwok
- Then check if the network have blockexplorer return it
- else use sepolia. 


In this case you have to think about the priority. 

I think you can can configure a custom netwoks with a blockexplorer.
But With your code if you have a blockexplorer it is using sepolia netwok.

hermannleboss (changes requested):

app/src/constant/network.ts
aliassan: It's following the functionality that you're looking for, you can run some tests to verify like this:

1. Try not setting alias and custom network
2. Try setting alias only
3. Try setting custom network only
4. Try setting both alias and custom network (If you set both then it will use alias, if you want custom to take precedence then we can change this)

The order of priority is like this:

1. Alias Network
2. Custom Network
3. Sepolia Network

It doesn't disallow to set custom network with Block Explorer URL, it just makes it optional like in MetaMask. If you set Block Explorer URL or don't set it your custom network will still be registered. 

The `checkNetwork` function just checks to make sure that all of the required network fields are set, `blockExplorer` is optional. If any of the required fields is missing it returns false and default network is used.

The best way is for you to test the functionality.

hermannleboss: Try to set only the alias in your .env and see what happen 

VITE_APP_NETWORK_ALIAS=

I made somes test it is not changing the network

hermannleboss: I check on my side i write etherem instead of ethereum for VITE_APP_NETWORK_ALIAS

hermannleboss (approved):

PR #98, opened 2024-05-12
feature: tips withdrawal amount

farrellh1:
# Description

We want the user can withdraw certain amount of the user's balance

Fixes # (#96)

## Type of change

# Checklist:

### DIFF
--- a/contract/contracts/Tips.sol
+++ b/contract/contracts/Tips.sol
@@ -76,18 +76,16 @@ contract Tips  is OwnableUpgradeable, ReentrancyGuardUpgradeable, PausableUpgrad
         emit SendTip(msg.sender, _teamMembersAddresses, msg.value, amountPerAddress);
     }
 
-    // TODO: Protection for reentrancy attack
-    function withdraw() external nonReentrant whenNotPaused{
+    function withdraw(uint256 amount) external nonReentrant whenNotPaused{
         uint256 senderBalance = balance[msg.sender];
         require(senderBalance > 0, "No tips to withdraw.");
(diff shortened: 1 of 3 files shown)


contract/test/TipsProxyt.spec.ts
```diff
@@ -2,7 +2,7 @@ import { ethers, ignition } from 'hardhat'
 import { expect } from 'chai'
 
 import TipsModule from '../ignition/modules/ProxyModule'
-describe.only('Tips Proxy', function () {
+describe('Tips Proxy', function () {
   describe('Proxy interaction', async function () {
```
farrellh1, line 6: had to remove .only because it doesn't run the other tests

contract/contracts/Tips.sol
hermannleboss, line 1: I think we have to versioning the Contract we will deploy. 
Looking for a better way to do it.

contract/test/TipsProxyt.spec.ts
hermannleboss: LGTM

hermannleboss (commented):
I need to check some stuff on the upgradable contract

contract/contracts/Tips.sol
farrellh1: i was thinking that we write some changes on the contract code and then after finish, we can deploy this new changes contract and then change the pointer of implementation contract to this contract

we dont have to create new file like TipsV2. unless there is major changes

farrellh1:
this PR too old, and for now the feature is not needed now

PR #104, opened 2024-05-15
fix/ logout on token fetch failure

### LINKED ISSUE #102: bug: failed to fetch token but the app still work
## Desctiption

**Describe the bug**
When the app is unable to fetch the token from the backend the app should logout.
But in our case the app just throw an error in the console

**To Reproduce**

1. Go to App, Login
2. Close the backend server (or make unavailable)
3. The app don't have any way to check his token

**Expected behavior**
When the app is unable to fetch the token from the backend the app should logout.

**Screenshots**
image

dasarathg68:
# Description

Fixes #102 

## Type of change

# Checklist:

### DIFF
--- a/app/src/apis/memberApi.ts
+++ b/app/src/apis/memberApi.ts
@@ -2,7 +2,6 @@ import { type Member } from '@/types'
 import { AuthService } from '@/services/authService'
 import { BACKEND_URL } from '@/constant/index'
 import { isAddress } from 'ethers' // ethers v6
-import { AuthAPI } from './authApi'
 
 interface MemberAPI {
   deleteMember(id: string): Promise<void>
@@ -11,7 +10,10 @@ interface MemberAPI {
 }
 export class FetchMemberAPI implements MemberAPI {
   async createMembers(newMembers: Partial<Member>[], id: string): Promise<Member[]> {
-    const token: any = AuthService.getToken()
+    const token: string | null = AuthService.getToken()
+    if (!token) {
+      throw new Error('Token is null')
+    }
 
     const requestOptions = {
       method: 'POST',
@@ -31,7 +33,6 @@ export class FetchMemberAPI implements MemberAPI {
     const response = await fetch(`${BACKEND_URL}/api/member/${id}`, requestOptions)
     const resObj = await response.json()
     if (response.status === 401) {
-      await AuthAPI.verifyToken(token)
       throw new Error('Unauthorized')
     }
     if (!resObj.success) {
@@ -40,7 +41,11 @@ export class FetchMemberAPI implements MemberAPI {
     return resObj.members
   }
   async updateMember(member: Partial<Member>, id: string): Promise<Member> {
-    const token: any = AuthService.getToken()
+    const token: string | null = AuthService.getToken()
+
+    if (!token) {
+      throw new Error('Token is null')
+    }
     const requestOptions = {
       method: 'PUT',
       headers: {
@@ -60,7 +65,6 @@ export class FetchMemberAPI implements MemberAPI {
     const response = await fetch(`${BACKEND_URL}/api/member/${id}`, requestOptions)
     const resObj = await response.json()
     if (response.status === 401) {
-      await AuthAPI.verifyToken(token)
       throw new Error('Unauthorized')
     }
     if (!resObj.success) {
@@ -69,7 +73,11 @@ export class FetchMemberAPI implements MemberAPI {
     return resObj.member
   }
   async deleteMember(id: string): Promise<void> {
-    const token: any = AuthService.getToken()
+    const token: string | null = AuthService.getToken()
+
+    if (!token) {
+      throw new Error('Token is null')
+    }
     const requestOptions = {
       method: 'DELETE',
       headers: {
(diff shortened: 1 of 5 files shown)


hermannleboss:
I think i should'nt logout. 
If i'm right logout remove the token
So instead of removing the token the status isAuth should be false.
- And the user have an error of can't acces the backend
- The user can be locked out. He can't interact with the plateform. 
- For ex: the login page appear with the user account information and a logout button.
- Or just a PopUp with the status.

dasarathg68:
> I think i should'nt logout. If i'm right logout remove the token So instead of removing the token the status isAuth should be false.
> 
> * And the user have an error of can't acces the backend
> * The user can be locked out. He can't interact with the plateform.
> * For ex: the login page appear with the user account information and a logout button.
> * Or just a PopUp with the status.

Why do we have to tell the user that they can't access the backend? It will be hosted anyway right?
Only developers will face this error while setting up their environment and they can check the error that's shown in the console.
If token verification is failed by any means, they shouldn't have access to the platform which is done by logging out and clearing the token.

hermannleboss:
You are right. Normally there is no way the user can't acces the backend. 
But imagine the backend go down
Or the user don't have internet.
In this way the user have the app who is working (because it's a SPA), but can't interact with the backend. 
So i think he should have an explanation about what is happening. 

And i was saying that normally the token check should no be done every time. But we should check only if the token is valid in the frontend. And normally if we make a request that return a 403 error then we check the user authentication.

dasarathg68:
> You are right. Normally there is no way the user can't acces the backend. But imagine the backend go down Or the user don't have internet. In this way the user have the app who is working (because it's a SPA), but can't interact with the backend. So i think he should have an explanation about what is happening.
> 
> And i was saying that normally the token check should no be done every time. But we should check only if the token is valid in the frontend. And normally if we make a request that return a 403 error then we check the user authentication.



> You are right. Normally there is no way the user can't acces the backend. But imagine the backend go down Or the user don't have internet. In this way the user have the app who is working (because it's a SPA), but can't interact with the backend. So i think he should have an explanation about what is happening.
> 
> And i was saying that normally the token check should no be done every time. But we should check only if the token is valid in the frontend. And normally if we make a request that return a 403 error then we check the user authentication.

@hermannleboss Okay so now I've refactored so that the verifyToken is called only when there is an authorized response and does not logout the user when the backend is down.
Can you give ideas as to how to let the user know? There's already a toast that intimates whenever there is a failure.

app/src/apis/memberApi.ts
```diff
@@ -30,12 +31,13 @@ export class FetchMemberAPI implements MemberAPI {
     const response = await fetch(`${BACKEND_URL}/api/member/${id}`, requestOptions)
     const resObj = await response.json()
     if (!resObj.success) {
+      await AuthAPI.verifyToken(token)
       throw new Error(resObj.message)
     }
     return resObj.members
```
farrellh1, line 37: maybe for the next refactor: 

i think we can create something like composable for fetching data from API so we dont have to write verifyToken in each function where we fetch data from API, in the composable there is a check if the route is public we dont have to verify the token, and if it's protected route we need to verify the token before we fetch the data.

dasarathg68: We're not verifying it each time. Only when there is an unauthorized response from the backend we call the verifyToken method.

farrellh1: right, but we still have to define it in every function that will call our API, so we can implement DRY principle https://en.wikipedia.org/wiki/Don%27t_repeat_yourself


```typescript
export function useFetch(url: string) {
  // some other codes
  function fetchApi(url: string) {
    const res = await fetch(url)
    if (res.statusCode == 401) throw Unauthorize
  
    return res
  }
    // some other codes
}
```

```typescript
function findAllMember() {
  const { fetchApi } = useFetch()

  const res = await fetchApi(url)
  // no longer need to add check if it's unauthorize

  // rest of the code..
}

function findAllUser() {
  const { fetchApi } = useFetch()

  const res = await fetchApi(url)
  // no longer need to add check if it's unauthorize

  // rest of the code..
}
```

or u can use VueUse https://vueuse.org/core/useFetch/ but im not really sure about it

dasarathg68: Wow that sounds neat. I'll create a separate issue for this refactor.

dasarathg68: But I don't think a composable is appropriate in this case. There's no state that needs to be maintained. Based on this https://www.youtube.com/watch?v=N0QrFKBZuqA , this does not satisfy any of the below conditions for it to be a composable
- Your function uses other composable 
- Your function uses lifecycle hooks
- Your function has stateful logic - ref

farrellh1: we can add state like isFetching, isSuccess, error, data like in the useFetch in vueUse

hermannleboss (changes requested):
You don't need to verify the token on 401. Just throw the error. 

Where you handle the call when its an Unauthorized error logout.

` if (response.status === 401) {
      await AuthAPI.verifyToken(token)
      throw new Error('Unauthorized')
    }`

hermannleboss (approved):

PR #105, opened 2024-05-15
Refactor/tips toast composable hr

hermannleboss:
# Description

Using the same interface for interaction with the blockchain

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/app/src/composables/tips.ts
+++ b/app/src/composables/tips.ts
@@ -8,8 +8,22 @@ import { ref } from 'vue'
 
 const tipsService = new TipsService(EthersJsAdapter.getInstance())
 
-export function useTipsBalance() {
-  const balance = ref<string>('0')
+interface IContractReadFunction<T> {
+  isLoading: ref<boolean>
+  error: ref<any>
+  data: ref<T>
+  execute: () => Promise<void>
+}
+
+interface IContractTransactionFunction {
+  isLoading: ref<boolean>
+  error: ref<any>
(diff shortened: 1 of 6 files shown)


app/src/composables/tips.ts
```diff
@@ -4,12 +4,28 @@ import type { EventResult, TipsEventType } from '@/types'
...
+  execute: () => Promise<void>
+}
+
+interface IContractTransactionFunction {
+  isLoading: Ref<boolean>
+  isSuccess: Ref<boolean>
+  error: Ref<any>
+  transaction: Ref<any>
+  execute: (...args: any[]) => Promise<void>
+}
```
hermannleboss, line 25: @aliassan can you check this interface for me ?
I can merge them. But i don't knwo what is the best way to do it.

aliassan: You can define a generic type for the arguments of `execute` which allows for an empty object and use `data` for both read functions and transaction functions so it can be containing a transaction hash or some data retrieved from the contract. For example:

```
type GenericArgs<T> = T | undefined | {}

interface IContractFunction<T> {
    isLoading: Ref<boolean>
    error: Ref<any>
    data: Ref<any>
    execute(args: GenericArgs<T>)
}
```
So `data` will contain whatever you need for the call. You supply T as an object containing the arguments you want when you need them.

hermannleboss: OK LGTM.

PR #110, opened 2024-05-16
fix: metamask popup without doing any action

### LINKED ISSUE #78: perf: metamask request to sign into the app at the loading
### Acceptance criteria

- [ ] Metamask Pop up should appear only when the user click on SIWE button.

farrellh1:
# Description

fix metamask popup without doing any action

Fixes #78 

## Type of change

# Checklist:

### DIFF
--- a/app/src/App.vue
+++ b/app/src/App.vue
@@ -1,16 +1,23 @@
 <script setup lang="ts">
 import { RouterView } from 'vue-router'
+import { ref, watch, toRaw } from 'vue'
+import { storeToRefs } from 'pinia'
+import { isAddress } from 'ethers'
+import { useToastStore } from '@/stores/toast'
+import { useUserDataStore } from '@/stores/user'
+import { FetchUserAPI } from '@/apis/userApi'
+import { AuthService } from '@/services/authService'
+
 import Drawer from '@/components/TheDrawer.vue'
 import NavBar from '@/components/NavBar.vue'
 import NotificationToast from '@/components/NotificationToast.vue'
-import { useToastStore } from '@/stores/toast'
-import { ref, watch, toRaw } from 'vue'
-import { storeToRefs } from 'pinia'
-import { useUserDataStore } from '@/stores/user'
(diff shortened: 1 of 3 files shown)


dasarathg68 (commented):
https://github.com/globe-and-citizen/cnc-portal/issues/78 The acceptance criterion is that metamask request to connect the wallet should appear only when user click on SIWE button.
This definitely does solve the issue but I don't think it's the best way to do it.
I think it's part of a bigger problem where the NavBar is mounted in the login page even when it's not supposed to. If we solve that, I think this would also be rectified in the process.

farrellh1:
> #78 The acceptance criterion is that metamask request to connect the wallet should appear only when user click on SIWE button. This definitely does solve the issue but I don't think it's the best way to do it. I think it's part of a bigger problem where the NavBar is mounted in the login page even when it's not supposed to. If we solve that, I think this would also be rectified in the process.

yes agree, and route.path somehow always show root path '/', i still haven't figure that out
image

image

dasarathg68:
> > #78 The acceptance criterion is that metamask request to connect the wallet should appear only when user click on SIWE button. This definitely does solve the issue but I don't think it's the best way to do it. I think it's part of a bigger problem where the NavBar is mounted in the login page even when it's not supposed to. If we solve that, I think this would also be rectified in the process.
> 
> yes agree, and route.path somehow always show root path '/', i still haven't figure that out image
> 
> image

Yes I noticed it as well but when we print the route object the values look right as shown below
image

farrellh1:
when u print entire `route` it will shows `/login` but when u print `route.path` it returns `/`. so it doesn't come into the condition if path is `/login`
image
image

hermannleboss (approved):
LGTM

hermannleboss:
I fixed it. By checking if the user is auth before i mount the nav Bar

In the App.vue file, the route is / before it's change to /login
My fix is not using the route again to check if the nav should be mount or not

farrellh1:
@hermannleboss i think u forgot to load the balance

PR #172, opened 2024-06-09
feature: enhance get all users to be able to filters

farrellh1:
# Description
Enhance getAllUsers endpoint to be able query filters

## Type of change

# Checklist:


backend/src/controllers/userController.ts
```diff
@@ -103,15 +103,40 @@ export const updateUser = async (req: Request, res: Response) => {
   }
 };
 export const getAllUsers = async (req: Request, res: Response) => {
-  const { page, limit } = req.query;
+  const { page, limit, ownerAddress, query } = req.query;
```
dasarathg68, line 106: Use req.body.address 
Look at this PR to get an idea
https://github.com/globe-and-citizen/cnc-portal/pull/168

backend/src/controllers/userController.ts
```diff
@@ -160,3 +185,44 @@ export const searchUser = async (req: Request, res: Response) => {
     return errorResponse(500, error, res);
   }
 };
+
+export const getAllContractors = async (req: Request, res: Response) => {
+  const { ownerAddress } = req.params;
```
dasarathg68, line 190: same here. use req.body.address

dasarathg68 (changes requested):

backend/src/controllers/userController.ts
farrellh1: oh wait forgot to remove this

backend/src/controllers/userController.ts
```diff
@@ -103,15 +103,41 @@ export const updateUser = async (req: Request, res: Response) => {
   }
 };
 export const getAllUsers = async (req: Request, res: Response) => {
-  const { page, limit } = req.query;
+  const { page, limit, query } = req.query;
+  const { ownerAddress } = req.body
```
dasarathg68, line 107: Hey, its supposed to be `const ownerAddress = req.body.address`
Look at the middleware below. JWT is signed with the address. jwt.verify() returns the address back to the payload which is then set to req.body so that we can access in any controller function
image

hermannleboss: @farrellh1 filter need to have validation. And be well documented. 
If you can filter by anything you want, i don't think it's welcome

farrellh1: it should be filtered by a key is that what you mean @hermannleboss ?

for example query params address will filter only for address, query params name will filter only by name

farrellh1:
cancelled

PR #192, opened 2024-06-14
Feat/notification system frontend

aliassan:
# Description

Create front-end for notification system.

Fixes # (issue)

## Type of change

# Checklist:


hermannleboss (changes requested):
Can you use the useCustomFetch composable to retrieve the data from the API

hermannleboss:
image

Why is the response have this structure ?

And i want to know why do we need to pass the user address as a query if we have the use address provided by the middleware  in the request

aliassan:
> image
> 
> Why is the response have this structure ?
> 
> And i want to know why do we need to pass the user address as a query if we have the use address provided by the middleware in the reques

aliassan:
> image
> 
> Why is the response have this structure ?
> 
> And i want to know why do we need to pass the user address as a query if we have the use address provided by the middleware in the request

I think the responses which return data will benefit from a more generic structure like:

```
res.status(201).json({
    data: {notification},
    success: true
})
```

A lot of handlers are returning a similar structure for data so maybe there can be a custom util for success response where you just pass the data object instead.

aliassan:
Yeah, you're right userAddress is not necessary, it's there in every request

app/src/apis/notificationApi.ts
dasarathg68, line 1: I suggest you use useCustomFetch in the view/component itself to handle the API calls as that is the convention Hermann and I decided to use for now as it is simpler.

aliassan: Yeah, I updated to use useCustomFetch

hermannleboss: I think you plan to remove this file

app/src/composables/useNotification.ts
hermannleboss, line 1: I don't get why you try create useNotification composable. 
You can simply use useCustomFetch in the nav component to handle this.

hermannleboss (changes requested):
Try to make the notification system request simple so you can do it quickly.

app/src/composables/useNotification.ts
aliassan: It's just for organisation it looks like NavBar is pretty cluttered as it is already, I just it easier to navigate and debug when the logic is in a separate file

app/src/components/NavBar.vue
```diff
@@ -79,17 +104,67 @@
...
+  execute: executeUpdateNotifications
+  //data: _notifications
+} = useCustomFetch<NotificationResponse>('notification', {
+  immediate: false,
+  beforeFetch: async ({ options, url, cancel }) => {
+    //options.body = JSON.stringify({ id: id.value })
+    url += `?id=${id.value}`
+    return { options, url, cancel }
+  }
+})
```
hermannleboss, line 146: You can create a reactive variable 
const updateEndPoint=ref("")

```suggestion
} = useCustomFetch<NotificationResponse>(updateEndPoint, {immediate: false})
```

And when you update the value of update endpoint and call the executeUpdateNotification. 
The event will trigger.

app/src/components/NavBar.vue
```diff
@@ -79,17 +104,67 @@
...
+  await executeFetchNotifications()
+}
+
+watch(notifications, () => {
+  if (notifications.value) {
+    const { data } = notifications.value
+    const idx = data.findIndex((notification: any) => notification.isRead === false)
+    isUnread.value = idx > -1
+  }
+})
```
hermannleboss, line 163: I still don't get why you need this function ?

If you want to know if the user read all his notification
you can simply use a computed property.

app/src/components/NavBar.vue
```diff
@@ -79,17 +104,67 @@
...
 </script>
 
-<style scoped></style>
+<style scoped>
+.unread {
+  font-weight: bold;
+}
```
hermannleboss, line 169: There is a tailwind utility class for this

backend/src/routes/notificationRoute.ts
```diff
@@ -1,9 +1,10 @@
 import express from "express";
-import { getNotification } from '../controllers/notificationController';
+import { getNotification, updateNotification } from '../controllers/notificationController';
 import { authorizeUser } from "../middleware/authMiddleware";
 
 const notificationRoute = express.Router()
 
 notificationRoute.get('/', authorizeUser, getNotification)
+notificationRoute.put('/', authorizeUser, updateNotification)
```
hermannleboss, line 8:
```suggestion
notificationRoute.put('/:id', authorizeUser, updateNotification)
```

It will be simple like this. 

We have the same format like others update request.

hermannleboss (changes requested):
I give you somes sugestions.

hermannleboss (approved):

PR #206, opened 2024-06-20
Feat: Logger System

aliassan:
# Description

Logger that logs information in the console in dev mode.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/app/src/composables/useSiwe.ts
+++ b/app/src/composables/useSiwe.ts
@@ -49,8 +49,8 @@ async function siwe() {
 
     router.push('/teams')
   } catch (error: any) {
-    log.error(parseError(error), 'Some random error message', new Error('Example error'))
     isProcessing.value = false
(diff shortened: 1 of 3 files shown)


app/src/composables/useSiwe.ts
```diff
@@ -49,12 +49,9 @@ async function siwe() {
 
     router.push('/teams')
   } catch (error: any) {
+    log.error(parseError(error), 'Some random error message', new Error('Example error'))
```
hermannleboss, line 52: i don't get why you have 3 input for the error here. 
I think you do it just for demo.

hermannleboss (approved):

app/src/composables/useSiwe.ts
aliassan: Yeah it was just a demo, I'll remove that

PR #229, opened 2024-06-26
Refactor/ Make error more reactive

### LINKED ISSUE #214: Better Error Handeling for SIWE
## Description

useSIWE need to be refactor to handles in a better way different states.

We have to reply to theses question
- What happens if the first request Failed ?
- Handle When the second request Failed. 
- And handle differents request status. 

function siwe() should be inside useSiwe() to make it respect the pattern

In this video the first request failed. 
But the error is unable to read a value of nonce. It's not clear for the user. 

https://www.loom.com/share/3910510c6c0d41ed88dc428569730a03?sid=0f833298-7561-443d-a326-56b91fa5113d

## Acceptance Criteria
image

aliassan:
# Description

Change SIWE composable to get better error handling.

Fixes #214 

## Type of change

# Checklist:

### DIFF
--- a/app/src/composables/useCustomFetch.ts
+++ b/app/src/composables/useCustomFetch.ts
@@ -2,9 +2,8 @@ import { AuthService } from '@/services/authService'
 import { BACKEND_URL } from '@/constant/index'
 import { createFetch } from '@vueuse/core'
 import { ref } from 'vue'
-import { getFetchErrorMessage } from '@/utils'
 
-export const errorMessage = ref('')
+export const responseSatus = ref(0)
 
 export const useCustomFetch = createFetch({
   baseUrl: `${BACKEND_URL}/api/`,
@@ -23,10 +22,8 @@ export const useCustomFetch = createFetch({
     },
     onFetchError({ data, response, error }) {
       const status = response?.status ?? 0
-      // Store the error response status
-      errorMessage.value = status
-        ? `Error ${status}: ${getFetchErrorMessage(status)}`
-        : `Network Error: ${error?.message ?? 'An unknown error occurred'}`
+      responseSatus.value = status
+
       return { data, response, error }
     }
   }
--- a/app/src/composables/useCustomFetch.ts
+++ b/app/src/composables/useCustomFetch.ts
@@ -1,6 +1,10 @@
 import { AuthService } from '@/services/authService'
 import { BACKEND_URL } from '@/constant/index'
 import { createFetch } from '@vueuse/core'
+import { ref } from 'vue'
+import { getFetchErrorMessage } from '@/utils'
+
+export const errorMessage = ref('')
 
 export const useCustomFetch = createFetch({
   baseUrl: `${BACKEND_URL}/api/`,
@@ -16,6 +20,14 @@ export const useCustomFetch = createFetch({
         }
       }
       return { options }
+    },
+    onFetchError({ data, response, error }) {
+      const status = response?.status ?? 0
+      // Store the error response status
+      errorMessage.value = status
+        ? `Error ${status}: ${getFetchErrorMessage(status)}`
+        : `Network Error: ${error?.message ?? 'An unknown error occurred'}`
+      return { data, response, error }
     }
   }
 })
--- a/app/src/composables/useCustomFetch.ts
+++ b/app/src/composables/useCustomFetch.ts
@@ -1,9 +1,6 @@
 import { AuthService } from '@/services/authService'
 import { BACKEND_URL } from '@/constant/index'
 import { createFetch } from '@vueuse/core'
-import { ref } from 'vue'
-
-export const responseSatus = ref(0)
 
 export const useCustomFetch = createFetch({
   baseUrl: `${BACKEND_URL}/api/`,
@@ -19,12 +16,6 @@ export const useCustomFetch = createFetch({
         }
       }
       return { options }
-    },
-    onFetchError({ data, response, error }) {
-      const status = response?.status ?? 0
-      responseSatus.value = status
-
-      return { data, response, error }
     }
   }
 })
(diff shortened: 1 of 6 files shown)


hermannleboss (changes requested):
The idea is to handle all error type. And make them explicit. 
For the user & the dev, having an error there is an error is not explicit. 
Normaly there are differents error case

Error with fetch 
- Network error
- Backend error : Unknown error : 500
- Un Authorized error

Error in the client 
- An error is catched. then:
- SIWE or Metamask error. 
In this case you can check and return an appropriate response.

And at the end you can handle the taost and the log in the composable directly

aliassan:
Okay, I may need further clarification there now, maybe an example. In the implementation I have now all the errors are being handled, both the Network and Client errors. I also have an instance where I have a separate error for dev in log and user friendly one for toast, check lines 36 and 37 in `siweMessageCreatorAdapter.ts`

hermannleboss:
> Okay, I may need further clarification there now, maybe an example. In the implementation I have now all the errors are being handled, both the Network and Client errors. I also have an instance where I have a separate error for dev in log and user friendly one for toast, check lines 36 and 37 in `siweMessageCreatorAdapter.ts`

OK let say i want this message to popup: 

When there is a 
- network error :There is an network error
- 500 Error: An error occured try later
- 400 Error: You can't make this request
- Metmask Error: Request Rejected by the user.

That what i want to say when i say more explicit error.

app/src/composables/useCustomFetch.ts
```diff
@@ -16,6 +20,14 @@ export const useCustomFetch = createFetch({
...
       }
       return { options }
+    },
+    onFetchError({ data, response, error }) {
+      const status = response?.status ?? 0
+      // Store the error response status
+      errorMessage.value = status
+        ? `Error ${status}: ${getFetchErrorMessage(status)}`
+        : `Network Error: ${error?.message ?? 'An unknown error occurred'}`
+      return { data, response, error }
```
hermannleboss, line 30: You don't need to handle siwe error here. 

Can you look at useFetch, make an exemple to see what are the error the return ?

Create a simple project and make a fetch to an endpoint

- Request an url the not exit
- Request an endpoint that return 500 error 
- Request an endpoint that return 400 error

and look at what are the different between their response : {data, error}


Here, you are just changing the error message. 

What i want is to handle the error in the SIWE.
Like

When i get the error from useFetch (no need to change the error that useFetch return)
Then if the error is 
- 500 error display to the user :An error occured try later
- network error :There is an network error
- 400 Error: You can't make this request

And in the siwe if we have the error from Metmask return: Request Rejected by the user.


When i say show to the user i'm talking about the toast.

app/src/utils/errorUtil.ts
hermannleboss, line 1: This look good to me.

app/src/composables/useSiwe.ts
```diff
@@ -25,37 +23,53 @@ function createSiweMessageCreator(address: string, statement: string, nonce: str
...
+      const address = await ethersJsAdapter.getAddress()
+      const {
+        //isFetching: isFetchingNonce,
+        error: fetchError,
+        data: nonce
+        //execute: executeFetchNonce
+      } = await useCustomFetch<string>(`user/nonce/${address}`).get().json()
+
+      if (fetchError.value) throw new Error(errorMessage.value)
+
```
hermannleboss, line 42: You can handle the error here

by formatting the error here.
So if you have fetch Error, Just create the toast. with the formatted message.

app/src/views/LoginView.vue
```diff
@@ -60,7 +60,19 @@ import IconGoogle from '@/components/icons/IconGoogle.vue'
...
+
+const { addErrorToast } = useToastStore()
+const { isProcessing, siwe, error } = useSiwe()
+
+watch(error, (newValue) => {
+  if (newValue) {
+    log.error(newValue)
+    addErrorToast(newValue)
+  }
+})
```
hermannleboss, line 75: And in this case you don't need to handle the toast when there is an error her.

hermannleboss (changes requested):

app/src/composables/useSiwe.ts
aliassan: Isn't easier to just handle the error once (in the catch block) instead of every time you call code that can throw an error. In this you case just format the message and throw it to be handled in the catch block

app/src/composables/useCustomFetch.ts
aliassan: I think it's easier to respond to the status codes instead of the error response because 404 error produces something like this; `Unexpected token '<', "<!DOCTYPE "... is not valid JSON`

aliassan: 500 produces `Internal Server Error` and 401 produces `Unauthorized`

app/src/composables/useSiwe.ts
aliassan: Ooh okay, I get it now. What if we user the logger for that, like this. In that case if an error occurs you have a stack trace that shows:

```
export const useCustomFetch = createFetch({
  baseUrl: `${BACKEND_URL}/api/`,
  combination: 'chain',
  options: {
    async beforeFetch({ options }) {
      const token = AuthService.getToken()
      options = {
        mode: 'cors',
        headers: {
          'Content-Type': 'application/json',
          Authorization: `Bearer ${token}`
        }
      }
      return { options }
    },
    onFetchError({ data, response, error }) {
      const status = response?.status ?? 0
      responseSatus.value = status
      console.log('[useCustomFetch] error: ', error)
      //Log the error
      log.error(
        `${
          status
            ? `Error ${status}: ${getFetchErrorMessage(status)}`
            : `Network Error: ${error?.message ?? 'An unknown error occurred'}`
        }`
      )
      return { data, response, error }
    }
  }
})
```

hermannleboss: When we have an error it appear in the console, in the case of fetch error.
The request have two purpose. 

- Readability for the dev
- And user experience.

And about your question to handle the error once. There is another way to do it. And make the code more efficient.


You can simply put the useCustom fetch line outside the function, and call the execute inside. 
So you can create a compute state that will return a build error. And add a watch inside that show the toast on error.

aliassan: Okay, I don't quite understand the implementation that returns a build error also, `ethersJSAdapter` and `siweAuthService` can also throw errors. Do you want to have a try catch block around each of these?

hermannleboss (approved):
LGTM

PR #239, opened 2024-07-01
feature/governance-main-v0.1

dasarathg68:
# Description
image

Explanation : https://www.loom.com/share/8ae2e8219c8d4074bc7e9e4095fa9b35?sid=a40b1e99-e638-4ea6-8f7e-081bb24461be
Fixes # (issue)

## Type of change

# Checklist:


contract/contracts/Voting.sol
```diff
@@ -0,0 +1,107 @@
...
+
+    function initialize() public initializer {
+        __Ownable_init(msg.sender);
+    }
+
+   function addProposal(Types.Proposal calldata _proposal) public onlyOwner whenNotPaused {
+        proposals[proposalCount] = _proposal;
+        emit ProposalAdded(proposalCount, _proposal.title,_proposal.description);
+        proposalCount++;
+    }
```
farrellh1, line 27: @hermannleboss is it possible to pass Struct into an function argument?

contract/contracts/Types.sol
farrellh1, line 1: i think we can create new folder called voting, and then we can put this file into the folder

contract/contracts/Voting.sol
dasarathg68: Yes, you can. You can check how I'm calling it from the Voting.spec.ts
``` 
const proposal = {
          title: 'Proposal 1',
          description: 'Description of Proposal 1',
          draftedBy: await owner.getAddress(),
          isElection: false,
          votes: { yes: 0, no: 0, abstain: 0 },
          candidates,
          isActive: true,
          voters: [
            {
              name: 'Member 1',
              memberAddress: await member1.getAddress(),
              isVoted: false,
              isEligible: true
            },
            {
              name: 'Member 2',
              memberAddress: await member2.getAddress(),
              isVoted: false,
              isEligible: true
            }
          ]
        }

        await expect(votingProxy.addProposal(proposal))
          .to.emit(votingProxy, 'ProposalAdded')
          .withArgs(0, proposal.title, proposal.description)

File: contract/contracts/Voting/Voting.sol [hunk 2]
```diff
@@ -0,0 +1,107 @@
...
+        proposalCount++;
+    }
+
+    function getProposals() public view returns (Types.Proposal[] memory) {
+        Types.Proposal[] memory _proposals = new Types.Proposal[](proposalCount);
+        for (uint256 i = 0; i < proposalCount; i++) {
+            _proposals[i] = proposals[i];
+        }
+        return _proposals;
+    }
```
farrellh1, line 35: i think we dont need to do this, instead we can read from event `ProposalAdded`

contract/contracts/Voting/Voting.sol
```diff
@@ -0,0 +1,107 @@
...
+
+    function initialize() public initializer {
+        __Ownable_init(msg.sender);
+    }
+
+   function addProposal(Types.Proposal calldata _proposal) public onlyOwner whenNotPaused {
+        proposals[proposalCount] = _proposal;
+        emit ProposalAdded(proposalCount, _proposal.title,_proposal.description);
+        proposalCount++;
+    }
```
farrellh1, line 27:
```suggestion
   function addProposal(Types.Proposal calldata _proposal) public onlyOwner whenNotPaused {
        proposalCount++;
        proposals[proposalCount] = _proposal;
        emit ProposalAdded(proposalCount, _proposal.title,_proposal.description);
    }
```
it should be like this, so the id starts from 1

contract/contracts/Voting/Voting.sol
farrellh1: and instead of this function we create get one proposal by `proposalId`

contract/contracts/Voting/Types.sol
```diff
@@ -0,0 +1,38 @@
...
+    }   
+    struct Proposal{
+        string title;
+        string description;
+        string draftedBy;
+        bool isElection;
+        bool isActive;
+        DirectiveVotes votes;
+        Candidate[] candidates;
+        Member[] voters;
```
farrellh1, line 30: for voters i prefer to use mapping(address => boolean), so we dont need to do looping in `findVoter` so it will be constant time complexity O(1)

farrellh1, line 30: i just realized, is it possible that we remove the `DirectiveVotes votes`, so instead we just use `Candidate[] candidates` in candidates we can just create candidate with name "Yes", "No", "Abstain" (dynamic based on the user input) and can be more than 3 choices. and also we can differentiate between election or not based on the value `isElection`

it will looks like this, ignore the naming u can use any naming that you're feel better
```solidity
struct Proposal{
    string title;
    string description;
    string draftedBy;
    bool isElection;
    bool isActive;
    DirectiveVotes votes;
    Choice[] choices;
    mapping(address => bool) voters;
}

struct Choice {
    string name;
    address addressChoice;
    uint256 votes;
}
```

dasarathg68: The reason why I have avoided mappings is because **_Assignments to structs or arrays in storage do not work if they contain mappings_**

contract/contracts/Voting/Voting.sol
dasarathg68: Alright, done!

contract/contracts/Voting/Voting.sol
dasarathg68: That event is emitted only when a proposal is added, how do we retrieve all proposals to display?

contract/contracts/Voting/Voting.sol
dasarathg68: In Solidity, arrays do not automatically handle gaps in indices when directly assigning values. If you attempt to assign a value to proposals[1] without first assigning a value to proposals[0], you will encounter an error because the array does not yet have an element at index 1.

contract/contracts/Voting/Voting.sol
dasarathg68: I'm thinking we'll return all proposals and then apply a filter

contract/contracts/Voting/Types.sol
dasarathg68: As for the Choice struct, good idea but don't you think it is better to separate so that it follows a more generalized approach for the future?
As we add more types of Proposals like Dividend Approvals, ESOP and policy changes, it would be easy to create a structure like struct ESOP{} and then define it inside Proposal as a member.

dasarathg68: > did u mean like this? https://github.com/farrellh1/cerberus-vault/blob/55e42cbe53f7e03e1291709ff593e255d23df11d/contracts/contracts/CerberusVault.sol#L20

Oh wow. Let me try again then.

dasarathg68: image

dasarathg68: > did u mean like this? https://github.com/farrellh1/cerberus-vault/blob/55e42cbe53f7e03e1291709ff593e255d23df11d/contracts/contracts/CerberusVault.sol#L20

image
@farrellh1 great example but look at what happens when I use mapping inside a struct.

contract/contracts/Voting.sol
dasarathg68: great catch! in that case, we can manually set each of the properties and assign 0 initially during the creation of the proposal in the Solidity function:

```
proposals[proposalCount].title = _proposal.title
//..
//...
// and so on
proposal[proposalCount].votes={yes:0,no:0,abstain:0};
 ```

File: contract/contracts/Voting/Types.sol (hunk 4 above)
Comment (dasarathg68) [reply contract/contracts/Voting/Types.sol:30]: check dm

Comment (farrellh1) [reply contract/contracts/Voting/Types.sol:30]: ohh, dont pass full struct of the proposal, only pass some of important data

Comment (dasarathg68) [reply contract/contracts/Voting/Types.sol:30]: Also, we cant pass in function parameters as a struct if we have a mapping

**Comment by stravid87** [comment]
Nice work. I particularly liked the Loom video and the flow chart. My general feedback is, for the contract, storing so much information on chain makes me nervous as a potential scaling issue when then time comes to have thousands and then hopefully millions of users. For this first iteration of the CNC, I do not think memory optimization is our critical task therefore I'm not providing any specific suggestion. However, in the future, I suspect there are very clever way to store veracity of data, per say, rather than the data itself through such things as hashes, Merkle trees, etc.

Looking forward to seeing your backend hooked up to your UI/frontend!

File: contract/contracts/Voting/Types.sol [hunk 5]
```diff
@@ -0,0 +1,38 @@
...
+        address candidateAddress;
+        uint256 votes;
+    }
+    struct Member{
+        string name;
+        bool isEligible;
+        bool isVoted;
+        address memberAddress;
+    }   
+    struct Proposal{
```
stravid87, line 22: Can structs take on interfaces in solidity? That is can

'interface Proposal {}'
'struct Directive {} is Proposal'
'struct Election {} is Proposal'
So that a function could call
function addProposal(Types.Proposal calldata _proposal) public onlyOwner whenNotPaused {
and then the user could pass in either a Directive or Election?

dasarathg68: Ah no, turns out interfaces can only be used to define function prototypes that are implemented later on.

dasarathg68:
> Nice work. I particularly liked the Loom video and the flow chart. My general feedback is, for the contract, storing so much information on chain makes me nervous as a potential scaling issue when then time comes to have thousands and then hopefully millions of users. For this first iteration of the CNC, I do not think memory optimization is our critical task therefore I'm not providing any specific suggestion. However, in the future, I suspect there are very clever way to store veracity of data, per say, rather than the data itself through such things as hashes, Merkle trees, etc.
> 
> Looking forward to seeing your backend hooked up to your UI/frontend!

Thanks for the feedback, I have been searching for better ways to store the data but I did checkout a few instructors explaining big projects on YT, and they all said one thing: "Move logic offchain, store data onchain" so that it is effectively decentralized.

hermannleboss:
@dasarathg68 i won't merge this into develop. 
Let's follow the new branch structure and rules for PR

PR #259, opened 2024-07-12
Feat/ps/backend

aliassan:
# Description

Backend for the permissioning system.

Fixes # (issue)

## Type of change

# Checklist:


app/src/components/roles/AddRoleCategoryForm.vue
```diff
@@ -0,0 +1,218 @@
...
+      v-else
+    >
+      {{ isNew ? `Create` : `Update` }}
+    </button>
+    <button class="btn btn-active" @click="emits('closeModal')">Cancel</button>
+  </div>
+</template>
+<script setup lang="ts">
+import { ref, computed } from 'vue'
+import IconPlus from '@/components/icons/IconPlus.vue'
```
dasarathg68, line 154:
```suggestion
import { PlusCircleIcon } from '@heroicons/vue/24/outline'
```

app/src/components/roles/AddRoleCategoryForm.vue
```diff
@@ -0,0 +1,218 @@
...
+
+    <div v-if="props.isNew">
+      <!--Roles-->
+      <section class="mt-10">
+        <div
+          v-for="(input, index) in roleCategory.roles"
+          :key="index"
+          class="collapse collapse-plus bg-base-200 border mt-2"
+        >
+          <input type="radio" name="my-accordion-3" />
```
dasarathg68, line 38: The accordions do not have the right behavior. (Does not toggle)
Use the AccordionComponent.vue component instead.

aliassan: Okay, I'll effect on a new request for the front-end

backend/src/controllers/role/entitlementTypeController.ts
```diff
@@ -0,0 +1,42 @@
+import { prisma, errorResponse } from "../../utils";
+import { Request, Response } from "express";
+
+const initEntTypes = [
```
hermannleboss, line 4:
```suggestion
const initEntitlementTypes = [
```
You should name it in a way it's easy to understand.

backend/src/utils/roleUtil.ts
hermannleboss, line 1: I've just remarks you utils functions.

Utils function they are calling prisma directly.
Utils function should only process a data and return it.
It should not have any side effect.

If you want the data in a format you can process it in utils function the return it so the contoller decide to call prisma

backend/src/routes/roleRoutes.ts
```diff
@@ -0,0 +1,26 @@
...
+
+roleRouter.post('/:id', createRole)
+roleRouter.put('/', updateRole)
+roleRouter.delete('/:id', deleteRole)
+roleRouter.post('/role-category', addRoleCategory)
+roleRouter.get('/role-category', getRoleCategories)
+roleRouter.get('/role-category/:id', getRoleCategory)
+roleRouter.put('/role-category', updateRoleCategory)
+roleRouter.delete('/role-category/:id', deleteRoleCategory)
+roleRouter.get('/entitlement-types', getEntitlementTypes)
```
hermannleboss, line 24: IF roles category are not related to role.
They should have their own routing fille.

hermannleboss (changes requested):
- [ ] Formatting
- [ ] Uitls

You are just doing the crud with no check if the user have the right to do it.
I think you will handle if the user can do somes stuff or not.

backend/src/utils/roleUtil.ts
aliassan: Those are actually CRUD for entitlements, the CRUD for entitlements happens within an API call to ether `role`  or `roleCategory` because they don't exist alone outside roles and role categories, so they have to call prisma because they do CRUD for entitlements. 

If I include the logic in the controllers for `role` and `roleCategory` controllers the code becomes too much and difficult to manage. Unless you're suggesting that I move them to their own controllers and not util.

hermannleboss (approved):
LGTM

backend/src/utils/roleUtil.ts
hermannleboss: For this we will try a new approach to fix the problem 
look to this issue
[#284](https://github.com/globe-and-citizen/cnc-portal/issues/284)

PR #260, opened 2024-07-13
fix: accordion v-model

farrellh1:
# Description

Fix accordion not follow how the radio button works 

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/app/src/components/__tests__/AccordionComponent.spec.ts
+++ b/app/src/components/__tests__/AccordionComponent.spec.ts
@@ -47,4 +47,22 @@ describe('AccordionComponent', () => {
       expect(wrapper.findAll('h1').length).toBe(3)
     })
   })
+
+  describe('Emits', () => {
+    it('emits update:modelValue', async () => {
+      await wrapper.findAll('input[type="radio"]').at(1)!.trigger('click')
+
+      expect(wrapper.emitted('update:modelValue')).toEqual([[1]])
+    })
+
+    it('only renders selected accordion', async () => {
+      const checkedIndex = 1
+      await wrapper.findAll('input[type="radio"]').at(checkedIndex)!.trigger('click')
+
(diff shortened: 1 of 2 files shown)


app/src/components/__tests__/AccordionComponent.spec.ts
hermannleboss, line 1: Hello @farrellh1 
About the test last time i was talking about how you can create test for a component where when you edit the component without changing the behaviour of your component, you should not need to update the test. 

In your case you are finding element 'input[type="radio"]' 

Let say in the future, we decide to change the radio input buy a button. And storing the variable of the active state. 

The behaviour don't change, but because we change the implementation you have also to change the test.

hermannleboss: For this you can use data-test for the selector.

look at this file https://github.com/globe-and-citizen/cnc-portal/blob/develop/app/src/components/forms/__tests__/AddMemberForm.spec.ts

hermannleboss (changes requested):

dasarathg68 (approved):
LGTM

app/src/components/__tests__/AccordionComponent.spec.ts
farrellh1: I see, but I haven't added the emits test for `AccordionComponent` before, it should be added in the accordion component PR but I forgot. so I added here

hermannleboss: Yes but now i'm requesting you refactor the test to follow the rule.

The selector you used for the test in not good and can lead into unnecessary error.

farrellh1: i applied your suggestion, please check thanks

hermannleboss (approved):

Issue #243, opened 2024-07-02
[Feature] Copy in the clipboard & Tooltip to see in etherscan. 


hermannleboss:
## Description

Everywhere we have an address, we should be able to coppy them to the clipboard and or display it in etherescan. 

User Story
When i hoover an address
Then I see a tooltip "See in polygonScan"
Then I can click on the link to see on polygonScan

## Acceptance Criteria

- [ ] Use of DaysiUI tooltip
- [ ] User Profile
- [ ] Team Bank Contract Address
- [ ] Team Member address

PR #266, opened 2024-07-17
feature: add tooltip, copy address, and see in block explorer

farrellh1:
# Description

Add tooltip and copy address feature

https://www.loom.com/share/e2898110f0ab44889c2a8b59cf47983e?sid=8630765f-1fc4-4576-a184-aca1bac9a086

Fixes # (#243)

## Type of change

# Checklist:

### DIFF
--- a/app/src/components/forms/__tests__/EditUserForm.spec.ts
+++ b/app/src/components/forms/__tests__/EditUserForm.spec.ts
@@ -1,0 +1,96 @@
+import { mount, type VueWrapper } from '@vue/test-utils'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import EditUserForm from '@/components/forms/EditUserForm.vue'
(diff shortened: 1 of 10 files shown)


farrellh1:
there is a problem(see picture _before_) with tooltip and modal from daisyui, its because of the `overflow-y-auto`, when I changed it to `overflow-y-visible`, there is a side effect (see picture side-effect). wdyt? is there any better solution?

before:
image 

after: 
image
with side effect:
image

app/src/components/forms/__tests__/EditUserForm.spec.ts
hermannleboss, line 1: Can you add more test to make this component reach the 100% of coverage ?

hermannleboss (commented):
LGTM
Can try to remove items that are defined but are never used ?

The are in the warning when you run npm run type-check

farrellh1:
btw, should i changed the UI of the members first to be like the one you suggestion @hermannleboss or just use this one?

PR #289, opened 2024-07-31
Feature/fix testing

hermannleboss:
# Description

Fix all testing waring and improve them

Fixes # (issue)

## Type of change

Testing warning

# Checklist:


app/src/components/TeamCard.vue
hermannleboss, line 1: The Warning error occurs because **useUserDataStore** is being called directly in the template, which is not allowed outside of the setup function. To fix this, you need to call **useUserDataStore** inside the setup function and then use the result in the template.

app/src/components/__tests__/BankTransactions.spec.ts
hermannleboss, line 1: Bank address is a require props.
I just add one to fix the issue.

app/src/components/__tests__/ProposalCard.spec.ts
hermannleboss, line 1: Adding router so the component can have access to route location.

app/src/views/__tests__/TeamView.spec.ts
hermannleboss, line 1: Setup must be use with precautions. 
I remove it here to fix the warning

dasarathg68: So we need to return variables from data(){} only?

scrip.bat
dasarathg68, line 1: What's this for?

script.sh
dasarathg68, line 1: This as well?

dasarathg68 (approved):

app/src/views/__tests__/TeamView.spec.ts
hermannleboss: Yes it can be done like that.
If you just need to setup the data. 
But some's ractive data will create error.

app/src/components/__tests__/ProposalCard.spec.ts
```diff
@@ -18,21 +25,21 @@ vi.mock('@/composables/voting', () => {
   return {
     useVoteElection: vi.fn(() => ({
       execute: vi.fn(),
-      isLoading: vi.fn(),
```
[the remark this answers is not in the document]
hermannleboss: Here the change is because isLoading is a **ref** an not a **function**.

scrip.bat
hermannleboss: I just create a script to remove local branch that are already merged

PR #335, opened 2024-09-09
test: add login fixture and add more test

farrellh1:
# Description

Add fixture login and add dashboard testing

Fixes # (issue)

## Type of change

# Checklist:


app/src/views/HomeView.vue
```diff
@@ -9,7 +9,7 @@
     </p>
 
     <button class="mt-6 btn btn-secondary">
-      <RouterLink to="/teams">Tip your Team</RouterLink>
+      <RouterLink to="/teams" data-testid="tip-your-team-button">Tip your Team</RouterLink>
```
hermannleboss, line 12: Why you use data-testid ?
isn't data-test enough?

app/test/e2e/dashboard.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+  beforeEach(async ({ page, login }) => {
+    await login()
+    await page.goto('http://localhost:5173')
+  })
+
+  test('should be able to access the dashboard', async ({ page }) => {
+    await expect(page.locator('h1').first()).toHaveText('Welcome To the CNC portal')
+    await expect(page.locator('p').first()).toHaveText(
+      'Our Website is still in construction. Look into our wonderfull tip feature and try to tip your team'
+    )
```
hermannleboss, line 18: No need to check all the text.

app/test/e2e/login.spec.ts
```diff
@@ -38,13 +38,8 @@ test.describe('Sign in', () => {
...
     // Click sign-in button
     await page.getByTestId('sign-in').click()
-    await page.waitForLoadState('networkidle')
-    // Switch network
-    // Note: default network metamask is Ethereum and the app default network is Sepolia
-    await metamask.approveSwitchNetwork()
 
-    // Wait for connect metamask popup to appear
-    await page.waitForTimeout(3000)
+    await page.waitForLoadState('networkidle')
```
hermannleboss, line 42: Comment this. 

What this is doing ?

I think you should test the case that you are changing the network. 
So you edit the .env parameter and you check the network switch.

app/test/fixtures/login.fixture.ts
```diff
@@ -0,0 +1,32 @@
...
+          })
+        })
+      })
+      await page.addInitScript(() => {
+        window.localStorage.setItem('isAuth', 'true')
+        window.localStorage.setItem('name', 'test')
+        window.localStorage.setItem('ownerAddress', address)
+        window.localStorage.setItem('nonce', '41vj7bz5Ow8oT5xaE')
+        window.localStorage.setItem('authToken', '1234567890')
+      })
```
hermannleboss, line 25: This is not a good idea. 

You are setting the value you expected. 

I think you should test the page like you don't know how the page is coded. 
And you know how the user should interact with it. 
Now you create the test like a user who is interacting with the page. 

What you can do is intercept the backend request but let the page do what he have to do himself.

hermannleboss (changes requested):
I let comment in files.

app/src/views/HomeView.vue
farrellh1: Oops i think i dont need to define that, I choose to use data-testid because there is a function in playwright to get by data-test-Id https://playwright.dev/docs/other-locators#id-data-testid-data-test-id-data-test-selectors, but I haven't try if we define it data-test

app/test/e2e/dashboard.spec.ts
farrellh1: does we only test all the action inside it?

hermannleboss (approved):

PR #366, opened 2024-09-29
Feature/board of director ali

aliassan:
# Description

Add Expense Account to BoD.

Fixes # (issue)

## Type of change

# Checklist:


farrellh1:
I think you need to handle in the expense tab where every interaction that had modifier onlyOwner (which means only bod can call) will add an action
you can see it on the bank tab, I handled it. for example, in bank, the flow will be like

**BOD**
- click transfer/pause/transfer ownership/push tip will open up a modal (maybe add a red text to info the bod this will create a bod action)
- add all the fields that are required for the bank interaction + u need to add a description because in bod action we need a description of why we do this such an action
- create an action to the body contract, encode the function signature, etc
- on success hit backend endpoint create action

aliassan:
> I think you need to handle in the expense tab where every interaction that had modifier onlyOwner (which means only bod can call) will add an action you can see it on the bank tab, I handled it. for example, in bank, the flow will be like
> 
> **BOD**
> 
> * click transfer/pause/transfer ownership/push tip will open up a modal (maybe add a red text to info the bod this will create a bod action)
> * add all the fields that are required for the bank interaction + u need to add a description because in bod action we need a description of why we do this such an action
> * create an action to the body contract, encode the function signature, etc
> * on success hit backend endpoint create action

Okay, this PR was to add expense account to the BOD tab, I'll create a separate PR for the expense account tab specifically

farrellh1 (approved):
alright, from me all good and i tested it and it works fine. but why do u comment some of the codes?

aliassan:
> alright, from me all good and i tested it and it works fine. but why do u comment some of the codes?

I commented out code for the Add Action modal because it was causing test coverage to fail and we're not using it in the UX for adding actions we're the existing UI of Bank Section and Expense Section so it looks like it's not necessary

app/src/components/sections/SingleTeamView/forms/AddActionForm.vue
hermannleboss, line 1: We was planing to stop using the form add action. 
So no need to use this form again.

hermannleboss (approved):
LGTM

PR #541, opened 2025-01-22
Refactor/siwe

### LINKED ISSUE #535: [Refactor] Remove `ether.js` : Refactore SIWE
## Description

Refactor SIWE in this way
SIWE should not depend on siweMessageCreatorAdapter
SIWE should not do the checking of wallet status. The check should be donne outside of siwe

And this syntax is not a good practice

Image

This need to be done before #534 

## Acceptance Criteria

- [ ] Do not use, SiweMessage Creatore Adpater, nor web3LibraryAdapter
- [ ] Remove `siweMessageCreatorAdapter.ts`
- [ ] Let the wallets checks to the App file / Check Composable
- [ ] Focus on only sign in with ethereum.
- [ ] Follow syntax guidelines

aliassan:
# Description

Refactor SIWE.

Fixes #535 

## Contribution

For your PR please add a commnent to eache file edited to explain the changes you made.

## Type of change

## Checklist

- /app

```bash
npm run build
npm run test:unit
npm run type-check
npm run lint
npm run format
```

- /backend

```bash
npm run build
npm run test
npm run lint
npm run format
```

- /contract

```bash
npm run test
npm run lint
npm run format
```

### DIFF
--- a/app/src/composables/useSiwe.ts
+++ b/app/src/composables/useSiwe.ts
@@ -9,71 +9,26 @@ import { useAccount, useSignMessage, useChainId } from '@wagmi/vue'
 import { SiweMessage } from 'siwe'
 import { useWalletChecks } from '@/composables'
 
-// function createSiweMessage(params: Partial<SiweMessage>) {
-//   // Create SiweMessage instance with provided data
-//   const siweMessage = new SiweMessage(params)
-//   // Call prepareMessage method to properly format the message
-//   return siweMessage.prepareMessage()
-// }
-
 export function useSiwe() {
-  const { addErrorToast } = useToastStore()
+  //#region Refs
   const authData = ref({ signature: '', message: '' })
   const apiEndpoint = ref<string>('')
+
+  //#region Composables
+  const { addErrorToast } = useToastStore()
+  const { setUserData, setAuthStatus } = useUserDataStore()
   const { address } = useAccount()
   const chainId = useChainId()
-  const { performChecks, isProcessing } = useWalletChecks()
-  const { setUserData, setAuthStatus } = useUserDataStore()
-
-  watch(address, async (newVal) => {
-    if (newVal) {
-      await executeSiwe()
-    }
-  })
-
   const { data: signature, error: signMessageError, signMessage } = useSignMessage()
+  const { performChecks, isProcessing } = useWalletChecks()
+  //#endregion
 
-  watch(signature, async (newVal) => {
-    if (newVal) {
-      authData.value.signature = newVal
-      await executeAddAuthData()
-      const token = siweData.value.accessToken
-      const storageToken = useStorage('authToken', token)
-      storageToken.value = token
-      apiEndpoint.value = `user/${address.value}`
-      await executeFetchUser()
-      if (!user.value) return
-      const userData: Partial<User> = user.value
-      setUserData(userData.name || '', userData.address || '', userData.nonce || '')
-      setAuthStatus(true)
-
-      router.push('/teams')
-
-      isProcessing.value = false
-    }
-  })
-
-  watch(signMessageError, (newVal) => {
-    if (newVal) {
-      log.error('signMessageError.value', newVal)
-      addErrorToast('Unable to sign SIWE message')
-      isProcessing.value = false
-    }
-  })
-
+  //#region useCustomeFetch
   const {
     error: siweError,
     data: siweData,
     execute: executeAddAuthData
   } = useCustomFetch<string>('auth/siwe', { immediate: false }).post(authData).json()
-
-  watch(siweError, (newVal) => {
-    if (newVal) {
-      log.info('siweError.value', newVal)
-      addErrorToast('Unable to authenticate with SIWE')
-      isProcessing.value = false
-    }
-  })
 
   const {
     error: fetchUserNonceError,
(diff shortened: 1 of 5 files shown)


app/src/composables/useSiwe.ts
```diff
@@ -1,44 +1,32 @@
...
-    if (newVal) {
-      addErrorToast(parseError(newVal))
-      log.error('connectError.value', newVal)
-      isProcessing.value = false
-    }
-  })
+  const { address } = useAccount()
+  const chainId = useChainId()
+  const { performChecks, isProcessing } = useWalletChecks()
+  const { setUserData, setAuthStatus } = useUserDataStore()
```
hermannleboss, line 26:
```suggestion
  const userDataStore= useUserDataStore()
```

app/src/composables/useSiwe.ts
```diff
@@ -56,12 +44,8 @@ export function useSiwe() {
...
       if (!user.value) return
       const userData: Partial<User> = user.value
-      useUserDataStore().setUserData(
-        userData.name || '',
-        userData.address || '',
-        userData.nonce || ''
-      )
-      useUserDataStore().setAuthStatus(true)
+      setUserData(userData.name || '', userData.address || '', userData.nonce || '')
+      setAuthStatus(true)
```
hermannleboss, line 48:
```suggestion
      userDataStore.setUserData(userData.name || '', userData.address || '', userData.nonce || '')
      userDataStore.setAuthStatus(true)
```

app/src/composables/useSiwe.ts
hermannleboss, line 1: // TODO, watch sinature: for newVal ==> use newSignature : Imporve readability
  // Add comment to explain the purpose of the watch, and every step of the process

// TODO groups all similar functions : useCustomFetch, watch and functions you create for the composable together
This order is better for readability: 
- Ref
- Composables
- Custom functions
- Watch

app/src/composables/useSiwe.ts
```diff
@@ -124,46 +108,29 @@ export function useSiwe() {
...
+        return
       }
 
-      if ((await metaMaskConnector.getChainId()) !== networkChainId) {
-        switchChain({
-          chainId: networkChainId,
-          connector: metaMaskConnector
-        })
+      if (address.value) {
+        await executeSiwe()
```
hermannleboss, line 132: You are not throwing an error so no need to have a try catch here. 

Normaly you sould have a ref that represent is the check is good or not. So it should not be the result of `await performChecks()`

await performChecks()
if(address.value && checkStatus){
        await executeSiwe()
}
isProcessing.value = false

// Then the watch will display error depending on what happen.

app/src/composables/useWalletChecks.ts
hermannleboss, line 1: https://wagmi.sh/vue/api/composables/useConnect#import

Check the documentation here. 
And the purpose of wallet check is to check if you have a wallet connected and to request you to change your network if you are not connected to the right one.

app/src/composables/useSiwe.ts
hermannleboss, line 1: Yes i've forgotten, improve the naming of your variables. By simply reading a variable without knowing his purpose make it diffuclt to understand. 
And type the return of fetch functions.

app/src/composables/useSiwe.ts
```diff
@@ -1,175 +1,149 @@
...
-      return
-    }
+    await performChecks()
+  }
+  //#endregion
+
+  //#region Watch
+  watch(isSuccessWalletCheck, async (newStatus) => {
+    if (newStatus) await executeSiwe()
+  })
```
hermannleboss, line 83:
```suggestion
    await performChecks()
    if (isSuccessWalletCheck) await executeSiwe()
  }
```

app/src/composables/useSiwe.ts
```diff
@@ -1,175 +1,149 @@
...
-      await executeSiwe()
-    }
-  })
+  //#endregion
 
+  //#region Composables
+  const { addErrorToast } = useToastStore()
+  const userDataStore = useUserDataStore()
+  const { address } = useAccount()
+  const chainId = useChainId()
```
hermannleboss, line 22: You should use signMessageAsync instead of signMessage. This allow you to wait for the signature.
const { data: signature, error: signMessageError, signMessageAsync} = useSignMessage()

hermannleboss: And by using this you won't need to use a watcher to watch the signature, you can simply process to the next step.

app/src/composables/useWalletChecks.ts
```diff
@@ -0,0 +1,100 @@
...
+      isSuccess.value = true
+    }
+  })
+
+  function resetRefs() {
+    isConnectRequired.value = false
+    isSuccess.value = false
+    isProcessing.value = false
+  }
+
```
hermannleboss, line 42: I don't understand why you need this function.
You should create computed property and it will be better.

app/src/composables/useWalletChecks.ts
hermannleboss: @aliassan 
Do not limit wallet check to meta mask only.
- order your function, variables, refs and watcher like for siwe. You need to do the same for every files you create. 
- the behavior of the check should be: 
  - Have a reactive state for if we have the wallet connected with the right network. 
  - Because we have a composable that check if the user is connected, we don't need it again
  - And then having a function to request network change when needed.

hermannleboss (commented):
There is some weird behavior
1- When you meta mask is installed, but not connected to the app, the the user try to sign in, he is requested to connect metamask but he is not requested to sign in with SIWE, then he is stuck. with the button fixed in loading state. 

2- When you have a wallet connected, and it's not meta mask you can't connect to the app, and you have the error saying meta mask is not installed. Normally it should work whatever is the wallet connected. 

By looking at the result of your code base, i think we should give to the user the ability to select his connector according to what wallet is installed in his browser.

hermannleboss (approved):
LGTM

PR #652, opened 2025-03-24
feat: graph for bank contract

### LINKED ISSUE #656: The graph schema

dasarathg68:
# Description

Fixes #656 


the-graph/subgraph.yaml
```diff
@@ -90,7 +90,51 @@ templates:
...
+        - Transfer
+        - Deposited
+        - TokenTransfer
+        - TokenDeposited
+        - PushTip
+        - PushTokenTip
+        - SendTip
+        - SendTokenTip
+        - TipsAddressChanged
+        - TokenAddressChanged
```
aliassan, line 113:
```suggestion
        - Transaction
```

the-graph/subgraph.yaml
```diff
@@ -90,7 +90,51 @@ templates:
...
+        - event: PushTokenTip(indexed address,address[],indexed address,uint256)
+          handler: handlePushTokenTip
+        - event: SendTip(indexed address,address[],uint256)
+          handler: handleSendTip
+        - event: SendTokenTip(indexed address,address[],indexed address,uint256)
+          handler: handleSendTokenTip
+        - event: TipsAddressChanged(indexed address,indexed address,indexed address)
+          handler: handleTipsAddressChanged
+        - event: TokenAddressChanged(indexed address,string,indexed address,indexed address)
+          handler: handleTokenAddressChanged
```
aliassan, line 137:
```suggestion
```

the-graph/src/bank.ts
```diff
@@ -0,0 +1,215 @@
...
+  Transfer,
+  Deposited,
+  TokenTransfer,
+  TokenDeposited,
+  PushTip,
+  PushTokenTip,
+  SendTip,
+  SendTokenTip,
+  TipsAddressChanged,
+  TokenAddressChanged,
```
aliassan, line 12:
```suggestion
  Transaction
```

the-graph/src/utils.ts
aliassan, line 1: I don't get why you added this file, it isn't necessary

the-graph/src/bank.ts
aliassan, line 1: Remove the other handlers and remain with:

1. handleDeposited
2. handleTransfer
3. handleTokenTransfer
4. handleTokenDeposited

For the implementation of the same see what I did in `the-graph/src/expenseAccountEip712.ts` Let's focus only on what's necessary for the Transfer History. We don't need to index everything.

aliassan (changes requested):
Now you need to make sure that a `Bank` template is created when a `ContractDeployed` event for a `Bank` contract is emitted in the `Officer` contract.

You can check how I'm deploying the dynamic `ExpenseAccountEIP712` when the `ContractDeployed` event is emitted in `handleContractDeployed` in `the-graph/src/officer.ts`.

Here are the steps:

1. Import the generated `Bank` template from generated templates
2. Edit the **if..else** in `handleContractDeployed` to also create the Bank data store template similar to Expense Account if `contractTyoe` event parameter is `Bank`

So it's supposed to be:

```ts
if (event.params.contractType === "ExpenseAccountEIP712") {
  //create an ExpenseAccountEIP712 template...
} else if (event.params.contractType === "Bank") {
  //create a Bank template... (Copy what I did for ExpenseAccountEIP712 above)
}
```

aliassan:
It's working but just clean up the handlers, entities and imports that are not used.

Issue #626, opened 2025-03-11
[Refactor] My Approved Expense Section


hermannleboss:
## Description

This component need to be refactored
- Use table component for the table
- Use team Store for to get the team
- Reload can be done by trigger fetchTeam using the team store
- TransferFromBankFrom Can be improved, to be independant : #627

Issue #595, opened 2025-02-28
[Refactor] Expense Account : `MyApprovedExpenseSection.vue` and approved address data exchange


hermannleboss:
## Description
This component need to be refactored
- Use table component for the table
- Use team Store for to get the team
- Reload can be done by trigger fetchTeam using the team store
- TransferFromBankFrom Can be improved, to be independant : #627 

TLDR; Use the store to share the data across theses components. 

The component `My Approved Expense` need to be updated when the component `Approved Addresses` update the address status. And to handle it well the best solution is to use the store for it.

## Acceptance Criteria

- [x] Use the table component
- [x] use the store to get the team
- [x] Update transfer From Bank
- [x] No search inside the component
- [x] Simplify the reload mechanism
- [x] Use the store to handle the reload on Address Approval Update

hermannleboss:
Need #713

PR #693, opened 2025-04-01
Refactor/my approved expense section

### LINKED ISSUE #713: [Backend] Update the ExpenseController
## Description

Update expense controller:

When adding an expense to a member, if the member has already a signed expense, we can't add a new one
When trying to update the status of the expense to "disable, expired, or limitReached", we will check by checking in the smart contract.
And we should have an endpoint that sync the status for the member.  

## Acceptance Criteria

- [ ] sync Status endpoint
- [ ] Add restriction

aliassan:
# Description

Refactor to use team store and reload.

Fixes #713, #626 and #595 

## Contribution

## Type of change

## Checklist

- /app

```bash
npm run build
npm run test:unit
npm run type-check
npm run lint
npm run format
```

- /backend

```bash
npm run build
npm run test
npm run lint
npm run format
```

- /contract

```bash
npm run test
npm run lint
npm run format
```


app/src/views/team/[id]/ExpenseAccountView.vue
```diff
@@ -5,19 +5,19 @@
...
+        teamStore.currentTeam?.teamContracts.find(
+          (contract) => contract.type === 'ExpenseAccountEIP712'
+        )
+      "
       :address="
-        team?.teamContracts.find((contract) => contract.type === 'ExpenseAccountEIP712')?.address ||
-        ''
+        teamStore.currentTeam?.teamContracts.find(
+          (contract) => contract.type === 'ExpenseAccountEIP712'
+        )?.address || ''
```
Comment (hermannleboss) [inline app/src/views/team/[id]/ExpenseAccountView.vue:16]: Create a computed property for the address, so it will be simple to get them

hermannleboss (commented):
First Review done. 
The expense API need to be updated.

I will work on it. 

Now focus more on the refactor of the component by keeping it simple. 
Update the template to use Table component.

ApprovedExpensesSection is still fetching the team directly

backend/src/controllers/teamController.ts
hermannleboss, line 1: Can you explain your update here ?
Normally the claim and wage deletion when you try to delete the team should work well. 
So i don't get your update

backend/src/controllers/expenseController.ts
```diff
@@ -33,15 +39,39 @@ export const addExpense = async (req: Request, res: Response) => {
...
+
+    const activeExpenses = syncedExpenses.filter((expense) => 
+      expense.status !== "expired" && expense.status !== "limit-reached"
+    )
+
+    // Check if the user already has an active expense
+    if (activeExpenses.length > 0) {
+      return errorResponse(400, "User already has an active expense", res);
+    }
+
```
hermannleboss, line 66: I think you don't need to sync the status of all expense. 
The rule is only one active expense. 

There will be some performances issue. 

If the user have thousand expense, we will need to get all expense, then sync on all. 

I think for the expense, when we create it the  we should simply save also the expiration date. So with this date we can just use the first filter to get all expense that are not expired. Then sync on the rest.

But i wont request a change here, i will create new issue for it.

app/src/stores/expenseStore.ts
```diff
@@ -0,0 +1,94 @@
...
+  // TODO: fetch teams on mounted
+  // Having caching system,
+  // And trigger fetch on demand
+  // Update cache system.
+
+  const userDataStore = useUserDataStore()
+  const route = useRoute()
+  const { addErrorToast } = useToastStore()
+
+  const allExpenseURI = ref<string>(`/expense?teamId=${route.params.id}`)
```
hermannleboss, line 19: Here you can simply use the team store current team ID to fetch the data instead of the route.

app/src/stores/expenseStore.ts
```diff
@@ -0,0 +1,94 @@
...
+    statusCode: allExpenseDataStatusCode
+  } = useCustomFetch(allExpenseURI, { immediate: false }).get().json()
+
+  const myApprovedExpenses = computed(() => {
+    if (allExpenseData.value) {
+      const expenses = allExpenseData.value.map(
+        (expense: { data: string; signature: `0x${string}` }) => ({
+          ...JSON.parse(expense.data),
+          signature: expense.signature
+        })
```
hermannleboss, line 35: Type directly the return of useCustomFetch so you don't have to type expense there.
It's a little bit tricky to understand here

app/src/stores/expenseStore.ts
```diff
@@ -0,0 +1,94 @@
...
+  })
+
+  const fetchAllExpenseData = async (teamId: string) => {
+    allExpenseURI.value = `/expense?teamId=${teamId}`
+    await executeFetchAllExpenseData()
+    return {
+      allExpenseData,
+      allExpenseDataError,
+      allExpenseDataIsFetching
+    }
```
hermannleboss, line 61: You don't need to return it here.
You already have the export in the store

app/src/components/sections/ExpenseAccountView/MyApprovedExpenseSection.vue
hermannleboss, line 1: The is a composable useContract Balance, you should use it to get the balance instead of doing it yourself. 
I will create an issue for it.

app/src/components/sections/ExpenseAccountView/ApprovedExpensesSection.vue
```diff
@@ -93,7 +87,7 @@ const {
 //#region Funtions
 const init = async () => {
```
hermannleboss, line 88: You can remove this function and call directly refetchExpenseAccountGetOwner
And you don't need to call in on mount. 
It's called by default.

hermannleboss (approved):
LGTM

Just somes few update, and fixe the issue. Just fixe the issue, no need to reach the coverage threshold

backend/src/controllers/expenseController.ts
aliassan: It's okay I can add the filter like I did when you sync for fetch it filters out all expired and limit-reached approvals from syncing.

app/src/stores/expenseStore.ts
aliassan: OK

hermannleboss (approved):
LGTM

PR #1265, opened 2025-10-20
[Bug] Native token approved expense transfer

achillegley:
# Description

## Intial Issue Description

Cannot transfer approved expense for native token because it gives insufficient balance error even when balance is available.

Fixes # (issue)

not an issues but i cnaged the position of the confirm approval to be displayed on right instead of left. 
image

update expenseUtils  to also calculate the spendable amount. 

## Contribution

## Type of change

## Contribution checklist


app/src/components/forms/TransferForm.vue
achillegley: Replaced the standalone selectedTokenId ref and its manual watcher with a computed getter/setter (and added computed to imports) so the select can bind directly to model.value.token.tokenId and automatically update model.value.token when the tokenId changes. 

Also added a watcher on props.tokens to initialize or reset the selected token if the tokens list changes, defaulting to the first token when the current selection is no longer present.

PR #1790, opened 2026-04-19
refactor(backend): Zod schema for Claim.fileAttachments + drop getContracts TODO

### LINKED ISSUE #1787: fix(backend): enforce team membership on GET /contract
Sub-issue of #1661 (item **B**). Deferred from #1786/#1785 per review direction.

## Problem

[contractController.ts:68](backend/src/controllers/contractController.ts#L68):

```ts
export const getContracts = async (req: Request, res: Response) => {
  const { teamId } = req.query as unknown as GetContractsQuery;

  try {
    // TODO restrict access to the team members
    const contracts = await prisma.teamContract.findMany({ where: { teamId } });
```

Any authenticated user can read any team's contracts by guessing `teamId`. The TODO has been in the codebase since the endpoint was first written.

## Fix

The `requireTeamMember` middleware factory landed in #1785 — apply it to the route:

```ts
// src/routes/contractRoutes.ts
contractRoutes.get(
  '/',
  validateQuery(getContractsQuerySchema),
  requireTeamMember('query.teamId'),
  getContracts
);
```

Then remove the TODO comment from `getContracts`.

## Acceptance

- `GET /contract?teamId=X` returns 403 when the caller is not a member of team `X`.
- `GET /contract?teamId=X` returns 403 when the team does not exist (non-member of a non-existent team).
- Existing member flows still return 200.
- Test added for the 403 path in `contractController.test.ts`.

### LINKED ISSUE #1789: refactor(backend): Zod schema for Claim.fileAttachments JSON field
Sub-issue of #1661 (item **H**).

## Problem

`Claim.fileAttachments Json?` stores `FileAttachmentData[]` (see [attachmentService.ts](backend/src/services/attachmentService.ts)). The TypeScript type exists, but there is no Zod schema enforcing the shape at write time.

Today the write path is constrained (`addClaimBodySchema` / `updateClaimBodySchema` validate the incoming `attachments` array), so malformed data is unlikely in practice. But the absence of an exported schema for the stored shape means:

- Read-side consumers have no single source of truth for the JSON structure.
- Future write paths (e.g. internal services, migrations) can skip validation without anyone noticing.
- The existing `req.body` attachment shape is duplicated across `addClaimBodySchema` and `updateClaimBodySchema` rather than referenced.

## Fix

1. Export a single `fileAttachmentSchema` / `fileAttachmentsSchema` from `src/validation/schemas/claim.ts` (or a dedicated `attachment.ts`).
2. Reference it from `addClaimBodySchema.attachments` and `updateClaimBodySchema.attachments` instead of duplicating the object shape.
3. Derive `FileAttachmentData` with `z.infer<typeof fileAttachmentSchema>` and update `services/attachmentService.ts` to use the inferred type.
4. In `refreshAttachmentUrls` / `deleteAttachments` (where the stored Prisma `Json` is cast to `FileAttachmentData[]`), parse with the schema and log/reject invalid records instead of blind-casting.

## Acceptance

- Single Zod schema defines the attachment shape; no duplication between add/update body schemas.
- `FileAttachmentData` is inferred from the schema.
- Read-side code parses `Claim.fileAttachments` through the schema before using it (or has a tested fallback path for legacy rows).
- Existing tests pass; new tests cover: (a) invalid attachment rejected on write, (b) legacy/malformed stored data handled gracefully on read.

hermannleboss:
Closes #1789 and closes #1787.

### Before

The `fileAttachments` object shape was duplicated inline in `addClaimBodySchema` and `updateClaimBodySchema`, and a separate `FileAttachmentData` TypeScript type lived in `services/attachmentService.ts`. Read paths cast blindly: `claim.fileAttachments as FileAttachmentData[]`.

### After

- **Single canonical schema** in [validation/schemas/claim.ts](backend/src/validation/schemas/claim.ts):
  ```ts
  export const fileAttachmentSchema = z.object({
    fileKey: z.string().min(1),
    fileUrl: z.string().url(),
    fileType: z.string().min(1),
    fileSize: z.number().positive(),
  });
  export const fileAttachmentsArraySchema = z.array(fileAttachmentSchema).max(10);
  export type FileAttachmentData = z.infer<typeof fileAttachmentSchema>;
  ```
- `addClaimBodySchema` / `updateClaimBodySchema` both reference `fileAttachmentsArraySchema` — no duplication.
- `services/attachmentService.ts` re-exports `FileAttachmentData` from the schema module for back-compat, and swaps the ad-hoc `typeof` guards for `fileAttachmentSchema.safeParse` in `refreshAttachmentUrls` and `deleteAttachments`.
- New `parseStoredAttachments(unknown): FileAttachmentData[]` — tolerantly parses a stored JSON column, dropping any entries that fail validation. Used by `updateClaim` to replace its `as FileAttachmentData[]` cast.
- Controllers drop their casts when calling `refreshAttachmentUrls` / `deleteAttachments`.

### Tests

New `validation/__tests__/claim.test.ts` covers the schema and `parseStoredAttachments`. Existing `attachmentService` tests updated to use valid URL strings (the schema now enforces `z.string().url()`; the old test fixtures used fake strings like `'old1'` and relied on the previous type-cast-only approach).

## #1787 — Drop `getContracts` authz TODO

Team contracts are intentionally public data; the restriction is no longer planned. The `// TODO restrict access to the team members` line is removed from `contractController.getContracts`. No behavior change.

## Verification

- `npm run lint` — clean
- `npx tsc --noEmit` — clean
- `npm test` — **539 passed** (+9 new), 10 skipped

## Test plan

- [ ] Uploading a claim with 10 attachments still works; 11 is rejected with the "Maximum 10 files" error.
- [ ] Editing a claim with pre-existing attachments still lets the user delete and add attachments; legacy/malformed stored entries are dropped silently on edit.
- [ ] `GET /claim` and `GET /weekly-claim` return refreshed presigned URLs for valid stored attachments and preserve malformed entries unchanged.
- [ ] `GET /contract?teamId=X` is callable by any authenticated user (public team data).

### DIFF
--- a/backend/src/controllers/__tests__/weeklyClaimController.test.ts
+++ b/backend/src/controllers/__tests__/weeklyClaimController.test.ts
@@ -341,7 +341,7 @@ describe('Weekly Claim Controller', () => {
       const memberAddress = '0x000000000000000000000000000000000000dEaD';
 
       mockGetPresignedDownloadUrl
-        .mockResolvedValueOnce('fresh-1')
+        .mockResolvedValueOnce('https://fresh-1.example.com')
         .mockRejectedValueOnce(new Error('presign failed'));
 
       const weeklyClaims = [
@@ -352,13 +352,20 @@ describe('Weekly Claim Controller', () => {
               id: 101,
               hoursWorked: undefined,
               fileAttachments: [
-                { fileKey: 'k1', fileUrl: 'old', fileType: 'image/png', fileSize: 1 },
+                {
+                  fileKey: 'k1',
+                  fileUrl: 'https://old.example.com',
+                  fileType: 'image/png',
+                  fileSize: 1,
+                },
               ],
             },
             {
               id: 102,
               hoursWorked: 2,
-              fileAttachments: [{ fileUrl: 'no-key', fileType: 'image/png', fileSize: 2 }],
+              fileAttachments: [
+                { fileUrl: 'https://no-key.example.com', fileType: 'image/png', fileSize: 2 },
+              ],
             },
             {
               id: 103,
@@ -369,7 +376,12 @@ describe('Weekly Claim Controller', () => {
               id: 104,
               hoursWorked: 4,
               fileAttachments: [
-                { fileKey: 'k2', fileUrl: 'old2', fileType: 'image/png', fileSize: 4 },
+                {
+                  fileKey: 'k2',
+                  fileUrl: 'https://old2.example.com',
+                  fileType: 'image/png',
+                  fileSize: 4,
+                },
               ],
             },
             {
@@ -398,8 +410,12 @@ describe('Weekly Claim Controller', () => {
       );
 
       expect(response.body[0].hoursWorked).toBe(10);
-      expect(response.body[0].claims[0].fileAttachments[0].fileUrl).toBe('fresh-1');
-      expect(response.body[0].claims[3].fileAttachments[0].fileUrl).toBe('old2');
+      expect(response.body[0].claims[0].fileAttachments[0].fileUrl).toBe(
+        'https://fresh-1.example.com'
+      );
+      expect(response.body[0].claims[3].fileAttachments[0].fileUrl).toBe(
+        'https://old2.example.com'
+      );
     });
 
     it('should return 200 for empty claims list', async () => {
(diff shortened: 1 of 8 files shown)


hermannleboss (commented):
LGTM

PR #1900, opened 2026-05-09
[Feat] Manual minting — Add %/Ending % toggle + allocation recap

### LINKED ISSUE #1869: [Feat] Manual minting — Add %/Ending % toggle + allocation recap
Image

## Manual minting — Add %/Ending % toggle + allocation recap

Parent goal: #1863
Sprint plan: #1862

Refines the "Issue tokens" modal so issuance always uses the **ending** ownership stake as the canonical input, with an explicit toggle for users who think additively, and a recap of the resulting allocation.

### Background

Today the `Ownership stake` field in the Issue tokens modal is ambiguous: there is no signal whether the entered percentage is the **target final stake** or an **additive delta**. The intended behavior is: the value entered is the recipient's stake **after** issuance — i.e. entering `23%` means the recipient ends up with exactly `23%`, not `currentStake + 23%`.

### Scope

- [ ] Lock baseline behavior: ownership stake field = **ending** percentage (final stake after issuance)
- [ ] Add a binary toggle on the modal:
  - [ ] **Add %** — additive, e.g. `+5%` on top of the recipient's current stake
  - [ ] **Ending %** — final stake after issuance (default)
- [ ] Add a recap line under the form summarizing the resulting allocation, e.g.:
  - `Issuing 64.21 M9W3Fv2 → recipient stake 28% (was 23%)`
  - `New total supply: 1655.21 M9W3Fv2`
- [ ] Both percentage and token-amount fields stay in sync with the selected mode

### Out of scope

- Add absolute amount mode (e.g. `+10` tokens) — dropped during sprint scoping; a binary toggle is enough.

achillegley:
# Description

## Intial Issue Description
Image

### Manual minting — Add %/Ending % toggle + allocation recap

Refines the "Issue tokens" modal so issuance always uses the **ending** ownership stake as the canonical input, with an explicit toggle for users who think additively, and a recap of the resulting allocation.

#### Background

Today the `Ownership stake` field in the Issue tokens modal is ambiguous: there is no signal whether the entered percentage is the **target final stake** or an **additive delta**. The intended behavior is: the value entered is the recipient's stake **after** issuance — i.e. entering `23%` means the recipient ends up with exactly `23%`, not `currentStake + 23%`.

#### Scope

- [ ] Lock baseline behavior: ownership stake field = **ending** percentage (final stake after issuance)
- [ ] Add a binary toggle on the modal:
  - [ ] **Add %** — additive, e.g. `+5%` on top of the recipient's current stake
  - [ ] **Ending %** — final stake after issuance (default)
- [ ] Add a recap line under the form summarizing the resulting allocation, e.g.:
  - `Issuing 64.21 M9W3Fv2 → recipient stake 28% (was 23%)`
  - `New total supply: 1655.21 M9W3Fv2`
- [ ] Both percentage and token-amount fields stay in sync with the selected mode

Fixes #1869 

If exit: The description of issues you find and fix in this PR.

Updated mintForm.vue to use the add and end stake. 

wrong config. 
image

good config. 
image

image

## Contribution

## Type of change

## Contribution checklist


app/src/components/sections/SherTokenView/forms/MintForm.vue
achillegley, line 1: Slimmed down orchestration; delegates stake mode/sync/recap logic to child section; submit state now driven by child validity/issued amount.

app/src/composables/investor/useMintStakeAllocation.ts
achillegley, line 1: introduced a dedicated composable to centralize reactive mint behavior. It handles mode-aware conversions, tracks which field was last edited, resyncs values when mode/recipient/supply changes, computes recap values, and enforces ending-mode validation consistently for both % and token amount entry.

app/src/utils/investorMintAllocation.ts
achillegley, line 1: Added pure utility functions for the allocation math and formatting layer. This includes stake/amount conversion formulas, issued-delta computation from input amount, non-rounding percentage truncation helper, and recap-line formatters so components stay thin and behavior is testable.

app/src/components/sections/SherTokenView/ShareholderList.vue
achillegley, line 1: Updated shareholder percentage display to use the shared formatter instead of integer division output, giving consistent two-decimal display and avoiding misleading whole-number truncation in the table.

achillegley, line 1: Switched percentage rendering to shared formatter with fixed 2-decimal display to align with on-chain-like precision expectations.
Concrete example

on our UI (total not up to  100%)

Image

the display onchain on polygon

Image

So the fix is to avoid truncate that can lead to gap. 

the  correction applied example. 

Image

the chain holders list on polygon

Image

app/src/composables/investor/useMintStakeAllocation.ts
hermannleboss, line 1: Having a so long composable is not a good practice.
Il faut isoler le composable. 
When you look at a compoable it should have only one responsibility
- Get a data A or update a data A
- There is no need to mixt bet data A, Get data b and updatate data C

hermannleboss (changes requested):
Need composable update

app/src/utils/investorMintAllocation.ts
```diff
@@ -0,0 +1,138 @@
...
+const TOKEN_DECIMALS = 6
+const PERCENTAGE_DISPLAY_DECIMALS = 2
+
+export const roundToDecimals = (value: number, decimals: number): number => {
+  return Math.round(value * 10 ** decimals) / 10 ** decimals
+}
+
+export const truncateToDecimals = (value: number, decimals: number): number => {
+  return Math.trunc(value * 10 ** decimals) / 10 ** decimals
+}
```
hermannleboss: Look for Intl.NumberFormat

app/src/composables/investor/useMintStakeAllocation.ts
```diff
@@ -0,0 +1,251 @@
...
+  roundToDecimals,
+  truncateToDecimals
+} from '@/utils/investorMintAllocation'
+import { type MintStakeFormState, type StakeMode } from '@/types/investor'
+
+const TOKEN_DECIMALS = 6
+const PERCENTAGE_DISPLAY_DECIMALS = 2
+type InputField = 'amount' | 'percentage'
+
+export function useMintStakeAllocation(
```
hermannleboss: This composable carries 4 unrelated responsibilities at once: derived reads, two-way input sync, validation, and recap string composition.

**Ask yourself**: if I removed this file, what's the *minimum* that genuinely needs to exist? A single `issuedAmount = computed(...)` in the component, plus a `watchEffect` to mirror the twin field, plus 4 `computed` recap strings, plus a zod `.refine()` on the schema for validation. No composable needed.

When 90% of a composable's exports are derivable from one core value, the composable is masking simplicity, not adding any.

app/src/composables/investor/useMintStakeAllocation.ts
```diff
@@ -0,0 +1,251 @@
...
+  })
+
+  const endingStakeValidationMessage = computed(() => {
+    if (!isEndingStakeInvalid.value) return null
+    const currentStake = truncateToDecimals(
+      currentStakePercentage.value,
+      PERCENTAGE_DISPLAY_DECIMALS
+    )
+    return `Ending % must be greater than recipient's current stake (${currentStake}%).`
+  })
```
hermannleboss: `UForm` + `z.refine()` with `path: ['amount']` already bubbles validation messages through `UFormField` natively. See `app/src/composables/useClaimForm.ts` (`.refine(... { path: ['hoursWorked'] })`) and `app/src/components/sections/ProposalsView/forms/CreateProposalForm.vue` for cross-field refinement that depends on other field values.

**Question**: is there a reason to render the error message manually instead of relying on the framework? If not, the refine pattern collapses 30+ lines into ~3.

app/src/composables/investor/useMintStakeAllocation.ts
```diff
@@ -0,0 +1,251 @@
...
+      recipientBalanceNumber.value
+    )
+    if (amount === null || isNaN(amount) || amount <= 0 || finalStake === null) return null
+
+    const currentStake = truncateToDecimals(
+      currentStakePercentage.value,
+      PERCENTAGE_DISPLAY_DECIMALS
+    )
+    const formattedFinalStake = roundToDecimals(finalStake, PERCENTAGE_DISPLAY_DECIMALS)
+    return `Recipient stake → ${formattedFinalStake}% (was ${currentStake}%)`
```
hermannleboss: The `was X%` is truncated (to match `ShareholderList`'s bigint division), but the `→ Y%` is rounded. After mint, `ShareholderList` will display the new stake **truncated** — so the recap can promise `→ 29%` and the list show `28.99%` a second later.

**Principle**: when a recap predicts a post-action state, format it identically to how that state will be displayed afterwards. Reuse `formatStakePercentageFromSupply` here too.

app/src/composables/investor/useMintStakeAllocation.ts
```diff
@@ -0,0 +1,251 @@
...
+      syncPercentageFromAmount()
+      return
+    }
+
+    if (state.percentage) {
+      syncAmountFromPercentage()
+    }
+  }
+
+  watch([() => state.address, () => state.stakeMode, totalSupplyRaw, recipientBalanceRaw], () => {
```
hermannleboss: `setStakeMode` writes `state.stakeMode = mode` and *then* explicitly calls `syncPercentageFromAmount()` / `syncAmountFromPercentage()`. The `watch` just below also depends on `state.stakeMode` and does the same sync at the next tick — so the sync runs twice.

**Question**: if you both mutate a watched value *and* call a function that reacts to it, which one is the source of truth? Pick one: either the watch handles everything (then `setStakeMode` is a one-liner), or the explicit calls handle everything (then `state.stakeMode` shouldn't be in the watch's deps).

app/src/composables/investor/useMintStakeAllocation.ts
```diff
@@ -0,0 +1,251 @@
...
+  recipientBalanceRaw: Ref<bigint | null | undefined>,
+  tokenSymbol: Ref<string | undefined>,
+  isRecipientAddressValid: Ref<boolean>
+) {
+  const lastEditedField = ref<InputField>('percentage')
+
+  const totalSupplyDisplay = computed(() => {
+    if (totalSupplyRaw.value === undefined || totalSupplyRaw.value === null) return null
+    return formatUnits(totalSupplyRaw.value, TOKEN_DECIMALS)
+  })
```
hermannleboss: (Nit) `totalSupplyDisplay` is exposed only so the template can render `Current supply: {{ totalSupplyDisplay }}`. That's a one-line `computed` in the component — no reason for the composable to own it.

app/src/composables/investor/useMintStakeAllocation.ts
```diff
@@ -0,0 +1,251 @@
...
+  const endingStakeValidationMessage = computed(() => {
+    if (!isEndingStakeInvalid.value) return null
+    const currentStake = truncateToDecimals(
+      currentStakePercentage.value,
+      PERCENTAGE_DISPLAY_DECIMALS
+    )
+    return `Ending % must be greater than recipient's current stake (${currentStake}%).`
+  })
+
+  return {
```
hermannleboss: Zero direct tests for this composable. All coverage routes through `MintForm.spec.ts`, even for the trickiest parts (`lastEditedField`, watch ordering, mode-dependent sync direction). When this breaks, debugging via a mounted component is much harder than via a focused composable spec.

**Principle**: if you give a piece of logic a name worth importing, give it a spec worth running in isolation. See `app/src/composables/investor/__tests__/useShareholderMigration.spec.ts` for the local pattern. (Best resolution: drop the composable per the first comment, in which case this concern goes away.)

hermannleboss (commented):
### High-level

The feature works and the math is correct, nicely done. My pushback is on **complexity that doesn't pay for itself** and on **conventions already established in the repo that this PR bypasses**. The principles to internalize before the next iteration:

1. **One source of truth** — for a mint form, the only value the contract actually needs is the **issued amount**. Everything else (percentage, recap, "was X% / → Y%") is *display* derived from `(issuedAmount, supply, balance, mode)`. When you have N pieces of state but only 1 is load-bearing, the other N-1 should be `computed`, not synced fields.

2. **The framework already does it** — Nuxt UI's `UForm` + zod `.refine()` renders validation messages under `UFormField` automatically. Writing a custom `isEndingStakeInvalid` flag + a manual `<p>` for the message + a guard in `onSubmit` reimplements three layers that already exist.

3. **Atomic composable convention** — `app/src/composables/investor/reads.ts` already exposes `useInvestorSymbol`, `useInvestorTotalSupply`, `useInvestorBalanceOf`. `writes.ts` already has `useIndividualMint` (commented, ready to enable). Inlining `useReadContract` / `useWriteContract` in the component bypasses the convention.

4. **A composable must describe what it does in one sentence** — `useMintStakeAllocation` does 4 unrelated things (derived reads + field sync + validation + recap copywriting). The name "Allocation" gives no hint of any of them. If you can't summarize the composable's job in one short sentence, the logic belongs somewhere else.

Concrete points are in the inline comments below.

app/src/components/sections/SherTokenView/forms/MintForm.vue
```diff
@@ -142,55 +205,67 @@ const { data: totalSupplyRaw } = useReadContract({
...
+  abi: INVESTOR_ABI,
+  address: investorsAddress,
+  functionName: 'balanceOf',
+  args: computed(() => [state.address as Address]),
+  query: {
+    enabled: computed(
+      () =>
+        !!investorsAddress.value && isAddress(investorsAddress.value) && isAddress(state.address)
+    )
+  }
```
hermannleboss: This new `useReadContract` for `balanceOf` (and the two existing `useReadContract` blocks just above for `symbol` and `totalSupply`, plus the inline `useWriteContract` further down) bypass atomic wrappers that already exist in the repo:

- `useInvestorSymbol()` — `app/src/composables/investor/reads.ts:25`
- `useInvestorTotalSupply()` — `app/src/composables/investor/reads.ts:35`
- `useInvestorBalanceOf(account)` — `app/src/composables/investor/reads.ts:55`
- `useIndividualMint` — `app/src/composables/investor/writes.ts:23` (commented, ready to uncomment)

**Principle**: before writing new contract IO, grep `composables/investor/` for the function name. The convention is "one composable = one contract call" and the rest of the codebase already follows it.

app/src/utils/investorMintAllocation.ts
```diff
@@ -0,0 +1,138 @@
...
+  if (finalStakePercentage === null) return ''
+
+  if (mode === 'ending') {
+    return String(roundToDecimals(finalStakePercentage, PERCENTAGE_DISPLAY_DECIMALS))
+  }
+
+  const deltaPercentage = finalStakePercentage - currentStakePercentage
+  if (deltaPercentage <= 0) return ''
+  return String(roundToDecimals(deltaPercentage, PERCENTAGE_DISPLAY_DECIMALS))
+}
```
hermannleboss: `computeAmountFromTargetStake`, `computeAmountFromPercentageInput`, `computeIssuedAmountFromAmountInput`, `computePercentageFromAmountInput`, `getFinalStakeFromAmount` — all five are rearrangements of one equation: `(balance + X) / (supply + X) = p`.

**Question**: if I have one equation viewed from different angles, do I need 5 named functions, or 1 equation + a few `computed`? The current API also mixes return types weirdly — `computeAmountFromPercentageInput` returns "issued OR ending balance depending on mode", which makes the name lie.

Most of these can be inlined as `computed` next to `issuedAmount`.

app/src/utils/investorMintAllocation.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export const formatStakePercentageFromSupply = (
```
hermannleboss, line 94: Two precision regimes coexist in this one file: `formatStakePercentageFromSupply` is bigint-pure (good — matches what's stored on-chain), while every other function above goes through `Number(formatUnits(...))` and loses precision above ~2^53.

**Question**: why is one display path safe and the rest of the math not? If float is intentional for the form (user input is float-y anyway), document it; otherwise pick one regime. The asymmetry is what makes this confusing to read.

app/src/components/sections/SherTokenView/forms/__tests__/MintForm.spec.ts
```diff
@@ -151,6 +162,89 @@ describe('MintForm.vue', () => {
...
+    const wrapper = mountForm({ memberInput: { name: 'Bob', address: VALID_ADDRESS } })
+
+    await wrapper.find('[data-test="add-mode-button"]').trigger('click')
+    await wrapper.find('[data-test="percentage-input"]').setValue('5')
+
+    const computedAmount = Number(
+      (wrapper.find('[data-test="amount-input"]').element as HTMLInputElement).value
+    )
+    expect(computedAmount).toBeGreaterThan(60)
+    expect(computedAmount).toBeLessThan(80)
```
hermannleboss: Expected value here is `(0.28 * 1000 - 230) / 0.72 ≈ 69.44`. The `[60, 80]` window lets a ±14% bug pass.

**Principle**: a test that passes with a wrong-but-close value doesn't catch the bug it was written for. Prefer `expect(computedAmount).toBeCloseTo(69.44, 1)`.

app/src/components/sections/SherTokenView/forms/MintStakeSection.vue
achillegley, line 1: Core logic hub for Add/Ending modes, range hinting, sync between % and shr, stake validation, ending-disabled-on-zero-supply behavior, and parent emits.

app/src/components/sections/SherTokenView/forms/MintRecapCard.vue
achillegley, line 1: Encapsulates recap rendering with blue/red states; now supports contextual validation message passthrough for placeholder/error recap.

app/src/components/sections/SherTokenView/forms/TwinAmountInputs.vue
achillegley, line 1: Pure input UI for paired %/shr fields with sync helper text and supply display; reads token symbol/supply via investor reads composables.

app/src/components/sections/SherTokenView/forms/__tests__/MintForm.spec.ts
achillegley: addressed

hermannleboss (approved):
LGTM

PR #1902, opened 2026-05-11
refactor: simplify safe composables using TanStack Query mutations

### LINKED ISSUE #1704: refactor: simplify safe composables using TanStack Query mutations
## Overview

The composables in `app/src/composables/safe/` currently manage their own loading/error state manually with `ref(false)` even though the project already uses **TanStack Vue Query v5**. This creates inconsistency and unnecessary boilerplate that TanStack mutations handle natively.

### 1. Manual loading/error state everywhere

Every action composable repeats the same boilerplate:

```ts
// useSafeApproval.ts, useSafeExecution.ts, useSafeTransfer.ts, etc.
const isApproving = ref(false)
const error = ref<string | null>(null)

async function approve() {
  isApproving.value = true
  error.value = null
  try {
    // ...
  } catch (e) {
    error.value = (e as Error).message
  } finally {
    isApproving.value = false
  }
}
```

TanStack `useMutation` already provides `isPending`, `isError`, `error`, `reset()` — making these refs redundant.

### 2. Mutations without real `mutationFn`

Several mutations in `safe.mutations.ts` exist **only for query invalidation** — their `mutationFn` is a no-op or nearly empty:

```ts
// useExecuteTransactionMutation, useUpdateSafeOwnersMutation, useDeploySafeMutation
mutationFn: async () => {},   // actual logic lives in the composable
onSuccess: () => queryClient.invalidateQueries(...)
```

This splits the logic unnaturally: the mutation does nothing, but the composable calls it just to trigger `onSuccess`. The mutation function should own the full operation.

### 3. Inconsistent patterns

- `useSafeTransactionConflicts` relies entirely on query state (no manual refs) ✅
- All action composables manage state manually ❌
- `SafeBalanceSection.vue` manually calls `queryClient.invalidateQueries()` after a transfer instead of letting the mutation's `onSuccess` handle it

### 4. Extra indirection layer

Components → Composable → Mutation → (no-op fn) → `onSuccess` invalidation

The composable layer adds little value when TanStack mutations already encapsulate async state + side effects. For simple cases, components could consume mutations directly.

### A. Move full operation logic into `mutationFn`

Each mutation should own the entire operation, not just the API call:

```ts
export function useApproveTransactionMutation() {
  const { loadSafe } = useSafeSdk()

  return useMutation({
    mutationFn: async ({ safeAddress, safeTxHash }: ApproveParams) => {
      const safe = await loadSafe(safeAddress)
      const signature = await safe.signHash(safeTxHash)
      await safeApiKit.confirmTransaction(safeTxHash, signature.data)
    },
    onSuccess: (_, { safeAddress }) => {
      queryClient.invalidateQueries({ queryKey: safeKeys.transactions(safeAddress) })
    },
  })
}
```

### B. Remove wrapper composables for simple cases

`useSafeApproval`, `useSafeExecution`, `useSafeTransfer` become thin wrappers with no real added value. Components can use the mutation directly:

```ts
// Before (component)
const { approve, isApproving } = useSafeApproval()

// After (component)
const { mutate: approve, isPending: isApproving } = useApproveTransactionMutation()
```

### C. Keep composables only for complex orchestration

Composables that compose multiple operations or hold non-trivial logic are still justified:

- **`useSafeOwnerManagement`** — batch transactions, propose vs. direct-execute branching
- **`useSafeTransactionConflicts`** — multi-query derived state
- **`useSafeSdk`** — SDK instance caching (keep as-is)

### D. Centralize `onSuccess` invalidation

Remove manual `queryClient.invalidateQueries()` calls from components (e.g., `SafeBalanceSection.vue`) and consolidate them in mutation `onSuccess` callbacks.

## Files Affected

| File | Action |
|------|--------|
| `composables/safe/useSafeApproval.ts` | Remove — replace with direct mutation usage |
| `composables/safe/useSafeExecution.ts` | Remove — replace with direct mutation usage |
| `composables/safe/useSafeTransfer.ts` | Simplify or remove |
| `composables/safe/useSafeDeployment.ts` | Simplify — move logic into mutation |
| `composables/safe/useSafeProposal.ts` | Evaluate — may still be justified as orchestration |
| `composables/safe/useSafeOwnerManagement.ts` | Keep — complex orchestration |
| `composables/safe/useSafeTransactionConflicts.ts` | Keep — already uses query pattern |
| `queries/safe.mutations.ts` | Move full operation logic into `mutationFn` |
| `components/sections/SafeView/**` | Update to consume mutations directly or simplified composables |

## Expected Benefits

- **Less code**: Remove ~150 lines of repeated loading/error boilerplate
- **Consistency**: All async state comes from TanStack (single source of truth)
- **Better DX**: `isPending`, `isError`, `reset()`, `onSettled` all work out of the box
- **Devtools**: Full TanStack Query devtools visibility into mutation state
- **Testability**: Easier to test mutations in isolation with `renderHook`

## Notes

- `useSafeSdk` caching logic should remain — it is SDK-level infrastructure, not a state-management concern
- The query key factory (`safeKeys`) is well-structured and should be kept as-is
- This refactor is purely internal — no user-visible behavior changes expected

achillegley:
# Description

## Intial Issue Description

The composables in `app/src/composables/safe/` currently manage their own loading/error state manually with `ref(false)` even though the project already uses **TanStack Vue Query v5**. This creates inconsistency and unnecessary boilerplate that TanStack mutations handle natively.

Fixes #1704  (issue)

If exit: The description of issues you find and fix in this PR.

- Removed redundant state management and error handling from useSafeTransfer, useSafeApproval, useSafeDeployment, and useSafeExecution composables.
- Introduced useTransferFromSafeMutation for handling token transfers from Safe, consolidating logic for ERC20 and native transfers.
- Updated useApproveTransactionMutation and useDeploySafeMutation to leverage centralized transaction service interactions.
- Enhanced error handling and toast notifications for improved user feedback.
- Deleted obsolete test file for useSafeProposal state.

## Contribution

## Type of change

- [ ] Perf review

## Contribution checklist


hermannleboss (commented):
## Review — Safe mutations refactor

Nice direction: removing the manual `ref(false)` boilerplate and pushing real logic into `mutationFn` matches the intent of #1704. The new `buildTokenTransferData` / `proposeOrExecuteSafeTransaction` utils are a good shape (pure functions, testable).

Before merging I'd like to discuss a few things — grouped below.

### 🔴 Likely bugs / regressions

1. **ERC20 balance invalidation regression** in `useTransferFromSafeMutation` — only the native balance key is invalidated; the previous `invalidateSafeBalances` also invalidated `readContract` keys per token. After a USDC transfer the UI will show stale balances.
2. **`createAddOwnerTx({ threshold: newThreshold })` when `newThreshold` is `undefined`** — `createRemoveOwnerTx` has a `|| currentThreshold` fallback but the add branch doesn't. Worth verifying what the Safe SDK does in that case.
3. **Hardcoded stablecoin decimals** in `buildTokenTransferData` — only `usdc`/`usdt`/`usdc.e` get 6 decimals, everything else falls back to `parseEther` (18). Any other ERC20 added later silently breaks.

### 🟡 Consistency / design

4. **Error handling style is mixed**: `SafeDeploymentCard` and `SafeTransactions` use `try/catch` around `mutateAsync`, while `RemoveOwnerButton` / `AddSignerModal` / `UpdateThresholdModal` / `SafeBalanceSection` use `mutate` + `onSuccess`/`onError`. Picking one keeps the codebase readable and lets us share an error formatter.
5. **Validation is imperative and duplicated** across `deploySafe`, `useUpdateSafeOwnersMutation`, `useApproveTransactionMutation`, `useExecuteTransactionMutation`, `validateSafeTransfer`. Zod is already used in the forms (`AddSignerModal`, `UpdateThresholdModal`) — exporting shared schemas from `types/safe.schemas.ts` would unify the two layers and remove most of the inline `if (!isAddress) throw` / `if (!safeTxHash) throw` lines. Also gives us conditional validation (e.g. "newThreshold required when ownersToAdd is non-empty") for free, which addresses point 2.
6. **Naming**: `useSafeSDK.deploySafe` doesn't actually deploy — it initializes a SDK instance with a `predictedSafe`. The real deployment happens in the mutation. `createPredictedSafeSdk` (or similar) would describe what it does.
7. **`proposeOrExecuteSafeTransaction`** does two distinct things behind a boolean — splitting into `proposeSafeTransaction` + `executeSafeTransaction` lets the call site say which one it wants without a flag.

### 🟡 Smaller things (typing / DRY / dead code)

8. Dead fields `safeTxHash` and `signature` in `UpdateSafeOwnersParams.body` are never read by the mutation — types should match reality.
9. `extractTransactionHash` / `waitForTransaction` use `any` with eslint-disable; the Safe SDK exposes types for that. Same logic is also inlined in `useExecuteTransactionMutation` — could just call the helpers.
10. `ZERO_ADDRESS` is redefined in `utils/safe.ts` while `zeroAddress` from viem is already imported in the same file.
11. `await import('@/utils/safe')` for `ZERO_ADDRESS` is a dynamic import inside a sync code path — looks like a circular-import workaround.
12. `['balance', { address, chainId }]` is the wagmi internal key shape; if wagmi changes it, invalidation silently breaks.
13. `/* eslint-disable max-lines */` on `SafeTransactions.vue` — debt we are inflicting on ourselves.

### 🟢 Tests

14. 7 specs deleted (~2.5k lines), 4 component specs added. Coverage went up but mainly because untested code disappeared. Would be reassuring to have a focused spec on the new mutation layer — at least one that mocks the SDK and exercises `proposeOrExecuteSafeTransaction` directly.

---

**Suggested path**: fix points 1, 2, 3, 4 in this PR; open a follow-up ticket for the Zod schema migration (point 5) + the typing/naming items (6–13). Happy to discuss.

app/src/queries/safe.mutations.ts
```diff
@@ -179,57 +231,159 @@ export function useExecuteTransactionMutation() {
...
+        value: string
+        data: string
+        operation: number
+      }> = []
+
+      // Add owners
+      for (const owner of ownersToAdd) {
+        const safeTx = await safeSdk.createAddOwnerTx({
+          ownerAddress: owner as Address,
+          threshold: newThreshold
```
hermannleboss: **Potential bug — `threshold: undefined`**

`createAddOwnerTx` is called with `threshold: newThreshold`. When the caller only adds owners without specifying `newThreshold` (which is the case for `AddSignerModal` with `shouldPropose: false`), this passes `threshold: undefined` to the Safe SDK.

The sibling `createRemoveOwnerTx` below has a `newThreshold || currentThreshold` fallback — why isn't the add branch protected the same way?

**Why it matters**: depending on SDK version, this can either throw a cryptic error or silently change the threshold. Either way the user-visible behaviour is unexpected.

**Suggestion**: either fall back to `currentThreshold` here too, or validate upstream that `newThreshold` is required when adding owners. A Zod `.refine` would express that cleanly:
```ts
.refine(
  ({ ownersToAdd, newThreshold }) =>
    !ownersToAdd.length || newThreshold !== undefined,
  { message: 'newThreshold required when adding owners', path: ['newThreshold'] }
)
```

app/src/queries/safe.mutations.ts
```diff
@@ -179,57 +231,159 @@ export function useExecuteTransactionMutation() {
...
+        signer: userStore.address!
+      })
+    },
+    onSuccess: (_, variables) => {
+      // Pending transactions (needed for proposals when threshold >= 2)
+      queryClient.invalidateQueries({ queryKey: safeKeys.transactions(variables.safeAddress) })
+      // Safe balance (always changes after transfer)
       queryClient.invalidateQueries({
-        queryKey: safeKeys.transactions(variables.pathParams.safeAddress)
+        queryKey: ['balance', { address: variables.safeAddress, chainId: chainId.value }]
```
hermannleboss: **Regression — ERC20 balances no longer invalidated**

The old `invalidateSafeBalances` in `SafeBalanceSection.vue` invalidated both the native balance key **and** the `readContract` key for every token address held by the Safe:
```ts
// before
await Promise.all(
  tokenAddresses.map((tokenAddress) =>
    queryClient.invalidateQueries({
      queryKey: ['readContract', { address: tokenAddress, args: [safeAddress], chainId }]
    })
  )
)
```
The new mutation only invalidates the native `['balance', ...]` key. After a USDC/USDT transfer, the displayed ERC20 balance will stay stale until the next route change.

**Question**: was this intentional? If not, we need to either invalidate the ERC20 `readContract` keys here too, or expose a `safeKeys.balance(safeAddress, tokenId?)` factory that owns the key shape so we don't depend on the wagmi internal.

hermannleboss: **Coupling to wagmi internal key shape**

`['balance', { address, chainId }]` is the internal shape wagmi uses for its balance query. If wagmi changes that (and they do across minor versions), this invalidation silently breaks with no test failure.

**Why it matters**: cache-key coupling is the kind of bug you only notice in prod when a user complains "my balance is wrong".

**Suggestion**: either import wagmi's key factory if exposed, or wrap it in a `safeKeys.balance(...)` helper next to `safeKeys.info` / `safeKeys.transactions` so we own the shape in one place.

app/src/queries/safe.mutations.ts
```diff
@@ -179,57 +231,159 @@ export function useExecuteTransactionMutation() {
...
+      }
+
+      // Use proposeOrExecuteSafeTransaction utility
+      return proposeOrExecuteSafeTransaction({
+        safeSdk,
+        transactionData,
+        shouldPropose,
+        chainId: chainId.value,
+        safeAddress,
+        signer: userStore.address!
```
hermannleboss: **Non-null assertion `userStore.address!`**

If the user disconnects their wallet between modal open and submit, this crashes with a non-descriptive `TypeError` instead of a friendly toast.

**Why it matters**: the rest of the app handles disconnection gracefully — this is a regression in UX.

**Question**: should `signer` be a required argument on the mutation params (validated by the schema), or should we read it lazily from `useConnection()` inside the mutation and throw a typed `WalletNotConnectedError`?

app/src/queries/safe.mutations.ts
```diff
@@ -150,23 +172,53 @@ export function useProposeTransactionMutation() {
...
+      const safeSdk = await loadSafe(safeAddress)
 
-  return useMutation<void, Error, ExecuteTransactionParams>({
-    mutationFn: async () => {
-      // Actual execution happens in the composable
-      // This mutation is just for query invalidation
+      // Transform and execute transaction
+      const sdkTransactionData = transformToSafeMultisigResponse(transactionData)
+      const txResponse = await safeSdk.executeTransaction(sdkTransactionData)
+
```
hermannleboss: **Duplicated logic with the helpers in `utils/safe.mutations.ts`**

This block manually extracts `txHash` and conditionally calls `wait()`, but `extractTransactionHash` and `waitForTransaction` already do exactly that.

**Why it matters**: when one of these is fixed (e.g. typing, error message), the other stays out of sync.

**Suggestion**: call the helpers here too, and fix their typing while we're at it (see comment on `utils/safe.mutations.ts`).

app/src/utils/safe.mutations.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (tokenAddress) {
+    return {
+      to: tokenAddress,
+      value: '0',
+      data: encodeFunctionData({
+        abi: erc20Abi,
+        functionName: 'transfer',
+        args: [
+          to as Address,
+          tokenId === 'usdc' || tokenId === 'usdt' || tokenId === 'usdc.e'
```
hermannleboss: **Hardcoded list of 6-decimal tokens**

```ts
tokenId === 'usdc' || tokenId === 'usdt' || tokenId === 'usdc.e'
  ? parseUnits(amount, 6)
  : parseEther(amount)
```
Any token added to the registry later that isn't one of these three (EURC, USDP, WBTC at 8 decimals…) will be transferred with the wrong unit. That's a fund-loss class of bug.

**Question**: do we have a token registry somewhere that exposes `decimals` per `tokenId`? If not, this is the right moment to introduce one — `getTokenDecimals(tokenId): number` next to the existing `getTokenAddress`.

app/src/utils/safe.mutations.ts
```diff
@@ -0,0 +1,143 @@
...
+    `${txServiceUrl}/api/v1/safes/${safeAddress}/multisig-transactions/`,
+    body
+  )
+}
+
+/**
+ * Extract transaction hash from Safe transaction response
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function extractTransactionHash(txResponse: any): string {
```
hermannleboss: **`any` + eslint-disable**

The Safe SDK exposes `TransactionResult` from `@safe-global/types-kit` — typing this as `TransactionResult` (or a narrower local type) removes the disable and surfaces shape changes at compile time.

**Question**: was the `any` a temporary shortcut, or did the SDK type not fit? If the latter, a small wrapper type would be better than disabling the lint rule.

app/src/utils/safe.mutations.ts
```diff
@@ -0,0 +1,143 @@
...
+export async function proposeOrExecuteSafeTransaction(params: {
+  safeSdk: Awaited<ReturnType<typeof import('@safe-global/protocol-kit').default.init>>
+  transactionData: Array<{ to: string; value: string; data: string; operation: number }>
+  shouldPropose: boolean
+  chainId: number
+  safeAddress: string
+  signer: Address
+}): Promise<string> {
+  const { safeSdk, transactionData, shouldPropose, chainId, safeAddress, signer } = params
+  const { ZERO_ADDRESS } = await import('@/utils/safe')
```
hermannleboss: **Dynamic import inside a hot path**

```ts
const { ZERO_ADDRESS } = await import('@/utils/safe')
```
The file already lives next to `utils/safe.ts` — this looks like a workaround for a circular import. Dynamic imports here add a microtask on every propose call and obscure the dependency graph.

**Question**: what's the cycle being broken? Often resolved by moving the constant to a leaf module (e.g. a `utils/safe.constants.ts`) that both files can import statically. Bonus: viem already exports `zeroAddress` so we might not need `ZERO_ADDRESS` at all here.

app/src/utils/safe.mutations.ts
```diff
@@ -0,0 +1,143 @@
...
+    operation: 0
+  }
+}
+
+/**
+ * Propose or execute Safe transaction based on threshold
+ */
+export async function proposeOrExecuteSafeTransaction(params: {
+  safeSdk: Awaited<ReturnType<typeof import('@safe-global/protocol-kit').default.init>>
+  transactionData: Array<{ to: string; value: string; data: string; operation: number }>
```
hermannleboss: **Two responsibilities behind a boolean flag**

`proposeOrExecuteSafeTransaction` branches on `shouldPropose` and has almost no shared code between the two branches (creating the transaction is the only common step).

**Why it matters**: every call site needs to know what the boolean means and read both branches to understand the behaviour. Hard to test one path in isolation.

**Question**: any objection to splitting into two named functions (`proposeSafeTransaction` / `executeSafeTransaction`) and letting `useUpdateSafeOwnersMutation` / `useTransferFromSafeMutation` pick one based on the threshold? Each function then has a single, clear contract.

app/src/utils/safe.ts
```diff
@@ -13,6 +13,20 @@ import { formatEtherUtil, tokenSymbol } from '@/utils/constantUtil'
 import { type DecodedCall } from '@/types'
 import type { SafeIncomingTransfer } from '@/types'
 
+export const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
```
hermannleboss: **`ZERO_ADDRESS` duplicates `zeroAddress` from viem**

The same file already imports `zeroAddress` from viem on line 1:
```ts
import { type Address, zeroAddress, formatUnits, isAddress } from 'viem'
```

**Question**: any reason not to re-export `zeroAddress` (or just use it directly at call sites)? Keeping two names for the same value invites inconsistency.

app/src/utils/safe.ts
```diff
@@ -200,3 +214,33 @@ export const formatSafeTransferType = (type: string): string => {
...
+
+/**
+ * Validate Safe transfer parameters
+ */
+export function validateSafeTransfer(options: SafeTransferOptions): ValidationResult {
+  if (!isAddress(options.to)) {
+    return { isValid: false, error: 'Invalid recipient address' }
+  }
+
+  if (!options.amount || parseFloat(options.amount) <= 0) {
```
hermannleboss, line 259: **`parseFloat` for amount validation loses precision**

```ts
if (!options.amount || parseFloat(options.amount) <= 0) { ... }
```
For amounts with more than 15 significant digits or scientific notation, `parseFloat` silently rounds.

**Why it matters**: the validation can pass a value that then fails in `parseUnits` downstream, or worse, transfers a different amount than the user typed.

**Suggestion**: validate via `parseUnits(amount, decimals)` and check the resulting `bigint > 0n`. A Zod schema (`AmountSchema`) shared with the `TransferForm` would centralize this.

app/src/types/safe.mutation.ts
```diff
@@ -46,16 +48,60 @@ export interface ExecuteTransactionParams {
...
+    /** Owners to remove */
+    ownersToRemove?: string[]
+    /** New threshold */
+    newThreshold?: number
+    /** Whether to propose the transaction */
+    shouldPropose?: boolean
+    /** Safe transaction hash */
+    safeTxHash?: string
+    /** Signature data */
+    signature?: SafeSignature | string
```
hermannleboss: **Dead fields in the params type**

`safeTxHash` (line 95) and `signature` (line 97) are declared on `UpdateSafeOwnersParams.body` but never read by `useUpdateSafeOwnersMutation` — it computes the signature itself via `safeSdk.signHash`.

**Why it matters**: dead fields lie to callers about what the mutation accepts; if someone passes them they'll have no effect and that's hard to debug.

**Suggestion**: drop them. If the schema is migrated to Zod (see PR-level comment), `z.infer` will keep types and runtime aligned automatically.

app/src/composables/safe/useSafeSdk.ts
```diff
@@ -61,9 +62,63 @@ export function useSafeSDK() {
...
 
+  /**
+   * Deploy a new Safe
+   * Initializes a Safe SDK instance with predicted safe configuration
+   *
+   * @param safeAccountConfig - Safe configuration (owners, threshold)
+   * @param deploymentConfig - Deployment configuration (saltNonce, safeVersion)
+   * @returns Safe SDK instance ready for deployment
+   */
+  const deploySafe = async (
```
hermannleboss: **`deploySafe` doesn't deploy**

This function initializes a SDK instance with a `predictedSafe` config — the actual on-chain deployment (`sendTransaction`) happens later in `useDeploySafeMutation`.

**Why it matters**: a future reader (or the next bug investigation) will assume `deploySafe` sends a transaction. Misleading names are a tax on every future change in this area.

**Suggestion**: rename to `createPredictedSafeSdk` or `initPredictedSafe`. The mutation can keep calling it but the name will tell the truth.

app/src/composables/safe/useSafeSdk.ts
```diff
@@ -61,9 +62,63 @@ export function useSafeSDK() {
...
+    if (!owners || owners.length === 0) {
+      throw new Error('At least one owner required')
+    }
+
+    if (threshold < 1 || threshold > owners.length) {
+      throw new Error(`Threshold must be between 1 and ${owners.length}`)
+    }
+
+    owners.forEach((owner, i) => {
+      if (!isAddress(owner)) {
```
hermannleboss, line 92: **Validation belongs in a schema**

The block from line 85 to 99 (owners non-empty, threshold in range, every owner is a valid address) is exactly the kind of input validation Zod is meant for. Zod is already used in the forms above (`AddSignerModal`, `UpdateThresholdModal`).

**Why it matters**: today this validation lives only at the SDK init layer, so the form layer re-validates the same things separately — drift becomes a question of when, not if.

**Suggestion**: define a `DeploySafeSchema` in `types/safe.schemas.ts`, parse at the entry of `useDeploySafeMutation`, and reuse the same schema in the form. This block then becomes a no-op.

app/src/components/sections/SafeView/SafeDeploymentCard.vue
```diff
@@ -94,22 +93,41 @@ const handleDeploySafe = async () => {
     return
   }
 
-  const safeAddress = await deploySafe([userDataStore.address!], 1)
-
-  if (!safeAddress) {
-    // Error already handled by deploySafe composable
+  let safeAddress: string
+  try {
```
hermannleboss: **`try/catch` around `mutateAsync` defeats TanStack state**

The rest of the codebase consumes mutations via `mutate(args, { onSuccess, onError })` — see `RemoveOwnerButton.vue`, `AddSignerModal.vue`, `UpdateThresholdModal.vue`, `SafeBalanceSection.vue`. Here we use `mutateAsync` + `try/catch`.

**Why it matters**: with `try/catch`, `isError` and `error` on the mutation never reflect the failure, `reset()` doesn't help, and the codebase has two styles for the same thing.

**Question**: was `mutateAsync` chosen to chain `createContract` afterwards? If so, you can still chain inside the first `onSuccess` (`createContract(...).then(...)` or another `mutate`). Picking the callback style throughout would let us share a `formatMutationError(err)` helper for the toast wording.

app/src/components/sections/SafeView/SafeTransactions.vue
```diff
@@ -305,8 +311,32 @@ const handleApproveTransaction = async (transaction: SafeTransaction) => {
   const safeAddress = props.address
   if (!safeAddress) return
   approvingTransactions.value.add(transaction.safeTxHash)
-  await approveTransaction(safeAddress, transaction.safeTxHash)
-  approvingTransactions.value.delete(transaction.safeTxHash)
+  try {
```
hermannleboss: **Same `try/catch` vs `onError` inconsistency**

Both `handleApproveTransaction` (here) and `handleExecuteTransaction` below use `mutateAsync` + `try/catch`. Other components in this PR use callbacks. Picking one style across the file (and the PR) makes the error wording reusable via a single helper.

**Question**: any reason `mutateAsync` was needed here, or can we move to `mutate(args, { onSuccess, onError })` like the modals?

app/src/components/sections/SafeView/SafeTransactions.vue
```diff
@@ -112,8 +112,9 @@
 </template>
 
 <script setup lang="ts">
+/* eslint-disable max-lines */
 import { watch, computed, ref } from 'vue'
```
hermannleboss, line 115: **`/* eslint-disable max-lines */`**

The file crosses the project's max-lines threshold and we disable the rule rather than splitting. The rule exists for a reason — once disabled, future additions land here too because "it already disabled the rule".

**Question**: can we extract the approval/execution handlers into a composable (e.g. `useTransactionRowActions`) so the rule stays green? Not blocking, but worth a follow-up issue if not done here.

app/src/components/sections/SafeView/SafeBalanceSection.vue
```diff
@@ -221,45 +220,42 @@ const resetTransferValues = () => {
...
+
+  transferFromSafe(
+    { safeAddress, options },
+    {
+      onSuccess: () => {
+        toast.add({
+          title: 'Success',
+          description: 'Transfer submitted successfully',
+          color: 'success'
+        })
```
hermannleboss, line 243: **`validateSafeTransfer` could live in the mutation**

This component calls `validateSafeTransfer` then `transferFromSafe` separately. The mutation could parse the same Zod schema internally, throw a `ZodError`, and the component would just need a single `onError` to format it.

**Why it matters**: today every call site of `useTransferFromSafeMutation` has to remember to call the validator first — and the next one will forget.

**Suggestion**: define `TransferFromSafeSchema` shared between this component's form and the mutation. The component-side `validateSafeTransfer` call goes away.

achillegley: Added shared transfer validation contract at type layer (transferFromSafeSchema) in app/src/types/safe.mutation.ts.  
Enforced that schema at mutation boundary (transferFromSafeSchema.parse({ safeAddress, options })) in app/src/queries/safe.mutations.ts (line 360).  
Removed duplicate component-side pre-validation (validateSafeTransfer) so transfer callers rely on mutation onError as single error path in
app/src/components/sections/SafeView/SafeBalanceSection.vue (line 223)

app/src/components/sections/SafeView/SafeDeploymentCard.vue
achillegley: Switched SafeDeploymentCard from mutateAsync + try/catch to callback-style mutation handlers:SafeDeploymentCard.vue (line 75)

deploySafe now uses mutate(..., { onSuccess, onError }).
createContract is chained inside deploy onSuccess, also using callbacks.

app/src/queries/safe.mutations.ts
achillegley: Added shared helper: walletUtil.ts (line 1)
Replaced local inline function in Safe mutation layer: safe.mutations.ts (line 21)
Reused same helper in Safe SDK composable (removed duplicate wallet checks): useSafeSdk.ts (line 6)

app/src/components/sections/SafeView/SafeTransactions.vue
achillegley: Switched mutation usage:
from mutateAsync to mutate for both approve/execute
Replaced try/catch/finally in:
handleApproveTransaction
handleExecuteTransaction

app/src/composables/safe/useSafeSdk.ts
achillegley: deploySafe was renamed to createPredictedSafeSdk to match actual behavior (SDK init only, no on-chain broadcast):
Renamed helper + docs in useSafeSdk.ts (line 64)
Updated deploy mutation call site in safe.mutations.ts (line 35)
Updated test names/usages in useSafeSdk.spec.ts (line 48)

app/src/composables/safe/useSafeSdk.ts
achillegley: Added shared Safe schemas in safe.schemas.ts:deploySafeSchema validates:owners is non-empty
every owner is a valid address
threshold is a positive integer and <= owners.length

moved transferFromSafeSchema there too for schema colocation.

Updated deploy mutation to parse at entry in safe.mutations.ts (line 40) via deploySafeSchema.parse(payload).
Moved deploy payload type to mutation types:added SafeDeploymentParams to safe.mutation.ts (line 3)
removed duplicate from safe.ts

Exported schemas through types/index.ts (line 23).

app/src/queries/safe.mutations.ts
achillegley: Refactored to remove duplicated tx-response handling and reuse shared helpers.
In app/src/utils/safe.mutations.ts (line 32), exported:extractTransactionHash
waitForTransaction

In app/src/queries/safe.mutations.ts (line 209), useExecuteTransactionMutation now uses those helpers instead of manually:reading txResponse.transactionResponse?.hash || txResponse.hash
conditionally calling wait()

app/src/queries/safe.mutations.ts
achillegley: is normaly optionnal. 
Addressed by enforcing newThreshold at the form/schema layer before mutation execution.
In app/src/components/sections/SafeView/forms/AddSignerModal.vue, added a Zod refine so adding owners requires newThreshold.
The submit payload now includes newThreshold explicitly when sending ownersToAdd to useUpdateSafeOwnersMutation.
Updated coverage in app/src/components/sections/SafeView/forms/__tests__/AddSignerModal.spec.ts.
Outcome: createAddOwnerTx no longer receives threshold: undefined

app/src/queries/safe.mutations.ts
achillegley: In safe.mutations.ts (line 390), useTransferFromSafeMutation.onSuccess now invalidates both:

native balance key: ['balance', { address, chainId }]
ERC20 keys per token: ['readContract', { address: tokenAddress, args: [safeAddress], chainId }]
So USDC/USDT balance invalidation is now included after transfer.
What was not added is a new safeKeys.balance(...) factory; current fix keeps the explicit query keys.

achillegley: Added Safe key factories in safe.queries.ts (line 35):safeKeys.balance(address, chainId)
safeKeys.tokenBalance(tokenAddress, safeAddress, chainId)

Updated transfer mutation invalidation in safe.mutations.ts (line 390) to use those helpers instead of inline:

app/src/types/safe.mutation.ts
achillegley: removed the dead fields from UpdateSafeOwnersParams.body in safe.mutation.ts (line 81):
removed safeTxHash?: string
removed signature?: SafeSignature | string
removed now-unused SafeSignature import

app/src/utils/safe.mutations.ts
achillegley: Added getTokenDecimals(tokenId) in app/src/utils/constantUtil.ts (line 39), next to getTokenAddress.
Updated Safe transfer encoding in app/src/utils/safe.mutations.ts (line 105) to use:parseUnits(amount, getTokenDecimals(tokenId))

app/src/utils/safe.mutations.ts
achillegley: Removed the boolean-flag API (proposeOrExecuteSafeTransaction) entirely.
Split into two explicit functions in safe.mutations.ts:proposeSafeTransaction (line 126 (line 126))
executeSafeTransaction (line 166 (line 166))

Updated call sites to choose explicitly:useUpdateSafeOwnersMutation (line 317 (line 317), line 326 (line 326))
useTransferFromSafeMutation (line 376 (line 376), line 385 (line 385))

So each function now has a single contract, and callers make the propose/execute decision directly.

app/src/utils/safe.ts
achillegley: Removed duplicate zero-address constant and standardized on viem’s zeroAddress.

Deleted ZERO_ADDRESS from app/src/utils/safe.ts.
Updated Safe proposal payload builder in app/src/utils/safe.mutations.ts to use zeroAddress directly for:
gasToken
refundReceiver

app/src/utils/safe.ts
achillegley: Replaced Number/parseFloat-based positive amount check in Safe transfer schema with shared parseUnits-based validation:app/src/types/safe.schemas.ts (line 35)

Reused the same shared positive-amount validation in TransferForm:app/src/components/forms/TransferForm.vue (line 202)

PR #1877, opened 2026-05-08
Feature/Add archive and hide functionality for teams

Georginio-prod:
# Description

## Intial Issue Description

This PR introduces the ability to archive and hide teams, enhancing team management and user experience.

Fixes #1304 

## Issues introduced and fixed (Optional)

If exit: The description of issues you find and fix in this PR.

## PR Summary Or Solution description

The summary should be a short description of the change, what you have done to fixe the issue. Please include the motivation for the change and how it was implemented. If applicable, please include screenshots or code snippets to illustrate the changes.

## Contribution

For your PR please add a comment to each file edited to explain the changes you made.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

- **Please check if the PR fulfills these requirements**

- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added / updated (for bug fixes / features)

## Contribution checklist

Before submitting the PR, please make sure you have applied the rules in [CONTRIBUTION.md](./../CONTRIBUTION.md)

PR #1904, opened 2026-05-11
docs: 📝 forbid duplicate query/mutation hooks for the same endpoint

### LINKED ISSUE #1903: docs: 📝 forbid duplicate query/mutation hooks for the same endpoint
## Context

When several UI actions hit the same backend endpoint, contributors sometimes create one mutation hook per action (e.g. `useArchiveTeamMutation`, `useUnarchiveTeamMutation`, `useHideTeamMutation`, `useShowTeamMutation`) even though they all share `method`, `endpoint`, and `invalidateKeys` — only the request body differs. The same trap exists on the read side (a separate `useGetArchivedTeamsQuery` instead of a `showArchived` query param on `useGetTeamsQuery`).

This bloats `app/src/queries/**`, multiplies the query-key surface, and forces components to import N hooks where one would do.

## Proposal

Add a rule to `AGENTS.md`: **one hook per endpoint, not per action.** If multiple actions share `method` + `endpoint`, reuse the single existing hook and vary the body / queryParams at the call site. Same rule for read hooks — add params instead of forking the hook.

### Pattern

```ts
const { mutate: updateTeam, reset } = useUpdateTeamMutation()

function archive() {
  updateTeam(
    { pathParams: { id: teamId }, body: { isArchived: true } },
    { onSuccess: () => reset() }
  )
}
function hide() {
  updateTeam(
    { pathParams: { id: teamId }, body: { isVisible: false } },
    { onSuccess: () => reset() }
  )
}
```

## Scope

- Applies to both `app/src/queries/**` mutation hooks and query hooks
- Add bullet under the existing `Frontend mutation pattern` / `Search before you create` rules in `AGENTS.md`
- Mention it explicitly so reviewers can flag it in PRs

## Acceptance

- [ ] `AGENTS.md` updated with the rule, linked from the existing mutation pattern section
- [ ] Wording covers both mutations and queries
- [ ] Short example included

hermannleboss:
## Summary

Adds an "One hook per endpoint, not per action" rule to `AGENTS.md` covering both mutations and queries in `app/src/queries/**`.

## Why

PR #1877 introduced `useArchiveTeamMutation`, `useUnarchiveTeamMutation`, `useHideTeamMutation`, `useShowTeamMutation` — four hooks identical to `useUpdateTeamMutation` except for the body. The rule makes the expected pattern explicit so reviewers can flag it consistently.

Closes #1903

## Test plan

- [ ] `AGENTS.md` reads correctly in GitHub preview
- [ ] Bullet placement makes sense next to existing `Frontend mutation pattern` rule

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -69,6 +69,19 @@ After cloning: `npm install` in each subproject you'll touch (`app/`, `backend/`
 
 **Frontend mutation pattern.** Mutations are a pure async function plus a `useXxxMutation` composable wrapping `@tanstack/vue-query`. Use `onSuccess` / `onError` callbacks rather than awaiting + try/catch in components. Surface errors reactively via `UAlert`. Issue #1776 documents the canonical pattern.
 
+**One hook per endpoint, not per action.** In `app/src/queries/**`, do not create multiple mutation or query hooks that share the same `method` + `endpoint` and differ only by request body or query params. Reuse the single existing hook and vary the input at the call site; on the read side, add a query param instead of forking the hook. Example: archive / unarchive / hide / show all reuse `useUpdateTeamMutation` — they do not need their own hooks. Reason: forking creates near-duplicate code, multiplies the query-key surface, and forces components to import N hooks where one would do. See issue #1903.
+
+```ts
+const { mutate: updateTeam, reset } = useUpdateTeamMutation()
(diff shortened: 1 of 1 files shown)


hermannleboss (commented):
LGTM

Issue #1735, opened 2026-03-29
[Perf] Full Migration to nuxt UI


hermannleboss:
## Description

The goals is to migrate all usage of daisy ui or other component into nuxt ui components

## Acceptance Criteria

- [ ] Remove DasyUI
- [ ] Remove vuepic

PR #1918, opened 2026-05-13
perf: ⚡️ replace @vuepic/vue-datepicker with custom UPopover month picker

### LINKED ISSUE #1702: [Perf] Remove @vuepic/vue-datepicker and use nuxt date picker
## Description

This is for form consistency

hermannleboss:
# Description

## Initial Issue Description

Fixes #1702 — [Perf] Remove `@vuepic/vue-datepicker` and use a Nuxt UI–native picker. The library was only used to drive the month/year picker inside `MonthSelector.vue`, so the whole dependency is dead weight.

Also progresses #1735 (full migration to Nuxt UI): closes the "Remove vuepic" acceptance criterion. The "Remove DaisyUI" criterion remains open and will be tracked in a follow-up.

## PR Summary

Replaces the `VueDatePicker` trigger in [MonthSelector.vue](app/src/components/MonthSelector.vue) with a self-contained `UPopover` panel:

- 3×4 month grid (Jan–Dec) built from `UButton`s; selected month gets `solid primary`, others `ghost neutral`.
- Year arrows above the grid (`← 2026 →`) instead of the library's separate year picker.
- The prev / next month side buttons keep the same wrap-around behavior they had before.
- The three call sites that built a `Week` from `(year, month)` are deduplicated into one `buildWeekFromMonth` helper, so the model contract (`isoString` = Monday of the ISO week containing the 1st, etc.) is preserved.

Then the dependency itself goes away:

- `@vuepic/vue-datepicker` removed from [package.json](app/package.json) and the lockfile.
- Global `app.component('VueDatePicker', …)` registration removed from [main.ts](app/src/main.ts).
- `import '@vuepic/vue-datepicker/dist/main.css'` removed from [App.vue](app/src/App.vue).

### Alternative considered

Reka UI's `MonthPickerRoot` (the primitive Nuxt UI inherits, currently alpha and not re-exported as `UMonthPicker`) was prototyped. It gives built-in arrow-key navigation and ARIA wiring but requires bridging `Week ↔ CalendarDate` and styling ~10 headless sub-components for a feature that only needs to pick a month. Net code is similar, surface area is larger, and the API is still alpha — so the custom popover wins here. Swapping it in later remains mechanical because the model logic is isolated in `buildWeekFromMonth`.

### Tests

[MonthSelector.spec.ts](app/src/components/__tests__/MonthSelector.spec.ts) is rewritten away from poking the old library's `setupState.monthPicked` and now drives the component through DOM clicks (`data-test="month-N"`, `data-test="next-year"`, `data-test="prev-month"`, …). Eight tests cover: picking a month from the grid, navigating the view year forward then picking, prev / next month with year wrap on both ends, the emitted `Week` shape, `viewYear` falling back to the current year when `modelValue` is undefined, and the parent-driven model.year watcher.

- `npm run test:unit` — 1659+/1659+ pass
- `npm run type-check` — clean
- `npm run lint` — clean

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) — removes a dependency without changing the UX contract
- [ ] New feature
- [ ] Breaking change

- [x] The commit message follows our guidelines (conventional commits + gitmoji, atomic)
- [x] Tests for the changes have been added
- [ ] Docs have been added / updated — n/a, no public-facing API change

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -26,7 +26,6 @@
         "@types/pdfmake": "^0.2.11",
         "@types/xlsx": "^0.0.35",
         "@vue/apollo-composable": "^4.2.1",
-        "@vuepic/vue-datepicker": "^11.0.3",
         "@vueuse/core": "^14.1.0",
         "@wagmi/core": "^3.0.0",
         "@wagmi/vue": "^0.4.3",
@@ -4722,16 +4721,6 @@
         }
       }
     },
-    "node_modules/@synthetixio/ethereum-wallet-mock/node_modules/zod": {
-      "version": "3.25.76",
-      "resolved": "https://registry.npmjs.org/zod/-/zod-3.25.76.tgz",
-      "integrity": "sha512-gzUt/qt81nXsFGKIFcC3YnfEAx5NkunCfnDlvuBSSFS02bcXu4Lmea0AFIUwbLWxWPx3d9p8S5QoaujKcNQxcQ==",
-      "extraneous": true,
-      "license": "MIT",
-      "funding": {
-        "url": "https://github.com/sponsors/colinhacks"
-      }
-    },
     "node_modules/@synthetixio/synpress": {
       "version": "4.1.2",
       "resolved": "https://registry.npmjs.org/@synthetixio/synpress/-/synpress-4.1.2.tgz",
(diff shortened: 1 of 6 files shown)


hermannleboss (commented):
LGTM

PR #1966, opened 2026-05-17
fix: 🐛 expense page critical UI issues (empty/error states, hardcoded values, a11y)

### LINKED ISSUE #1965: fix: 🐛 expense page critical UI issues (empty/error states, hardcoded values, a11y)
## Context

An audit of the Expense Account page (`ExpenseAccountView.vue`) surfaced 4 critical UI issues. This issue tracks fixing them in a single PR.

## 1. Hardcoded / fake values in stat cards

- `ExpenseAccountTotalApproved` displays a hardcoded `10` with a "TODO" subtitle → derive the real count from `useGetExpensesQuery`.
- `ExpenseMonthSpent` shows a fake "than last week" delta → compute a real delta, or drop it if no source.
- `ExpenseAccountBalance` shows a fake "than last week" delta → remove it (no cheap historical source).

## 2. Missing empty states

`MyApprovedExpenseSection`, `ExpenseAccountTable` and `ExpenseTransactions` render an empty table with no message. Add an empty state, and distinguish loading vs empty (skeleton/spinner while loading, message only when `!loading && rows === 0`).

## 3. Error states for failed fetches

`ExpenseTransactions` (and similar) only log fetch errors, so the user sees an empty table indistinguishable from "no data". Add an inline error state in the affected components. Mutation errors in forms/modals keep using `UAlert`.

## 4. Accessibility

- Add `aria-label` to modal dismiss buttons / titles.
- Status badges are differentiated only by color → add text/icon so they are color-blind accessible.

## Out of scope (separate tickets)

- Ponder query invalidation after a transfer (`ExpenseTransactions` / `ExpenseMonthSpent` not actively refreshed).
- Responsive / i18n issues.

hermannleboss:
## Summary

Fixes the 4 critical UI issues found while auditing the Expense Account page (`ExpenseAccountView.vue`).

### 1. Hardcoded / fake stat-card values
- `ExpenseAccountTotalApproved` — replaced the hardcoded `10` and the `+12.3% TODO` delta with the real count of distinct approved members from `useGetExpensesQuery`.
- `ExpenseMonthSpent` — replaced the fake `+26.3% than last week` with a real month-over-month delta computed from a previous-month transactions query. The delta is hidden when there is no previous-month baseline rather than faked.
- `ExpenseAccountBalance` — removed the fake `+41.3% than last week` delta (no historical balance source exists).

### 2. Empty states
`MyApprovedExpenseSection`, `ExpenseAccountTable` and `ExpenseTransactions` now show a contextual empty message instead of a blank table, and distinguish loading from empty.

### 3. Error states
`ExpenseAccountTable` and `ExpenseTransactions` show an inline error message when their fetch fails, so a failure is no longer indistinguishable from "no data". Mutation errors keep using `UAlert` as before.

### 4. Accessibility
- Status badges carry an icon so status is not conveyed by colour alone.
- The decorative `OverviewCard` icon is marked `aria-hidden` with an empty `alt`.

## Testing

- `npm run type-check` — passes.
- `npx vitest run` — 207 files / 1738 tests pass, plus new specs for `ExpenseAccountTotalApproved`, `ExpenseMonthSpent` and `MyApprovedExpenseSection`.
- `eslint` + `prettier` — clean on changed files.

## Out of scope (separate tickets)

- Ponder query invalidation after a transfer (`ExpenseTransactions` / `ExpenseMonthSpent` not actively refreshed).
- Responsive / i18n issues.

Closes #1965

### DIFF
--- a/app/src/components/OverviewCard.vue
+++ b/app/src/components/OverviewCard.vue
@@ -1,7 +1,13 @@
 <template>
   <div class="w-full rounded-2xl py-6" :class="[bgColor, textColor]" :data-variant="variant">
     <div class="flex flex-col items-center gap-4">
-      <img :src="cardIcon" alt="icon" class="h-16 w-16" data-test="card-icon" />
+      <img
+      :src="cardIcon"
+      alt=""
+      aria-hidden="true"
+      class="h-16 w-16"
+      data-test="card-icon"
+    />
       <span v-if="!loading" class="text-4xl font-bold" data-test="amount">{{ title }}</span>
       <USkeleton v-else class="h-8 w-24 opacity-20" />
       <span class="text-sm font-semibold" data-test="subtitle">{{ subtitle }}</span>
--- a/app/src/components/OverviewCard.vue
+++ b/app/src/components/OverviewCard.vue
@@ -1,13 +1,7 @@
 <template>
   <div class="w-full rounded-2xl py-6" :class="[bgColor, textColor]" :data-variant="variant">
     <div class="flex flex-col items-center gap-4">
-      <img
-      :src="cardIcon"
-      alt=""
-      aria-hidden="true"
-      class="h-16 w-16"
-      data-test="card-icon"
-    />
+      <img :src="cardIcon" alt="" aria-hidden="true" class="h-16 w-16" data-test="card-icon" />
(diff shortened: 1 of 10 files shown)


hermannleboss (commented):
LGTM

PR #1856, opened 2026-05-03
refactor: ♻️ migrate useSiwe to TanStack Query + Axios

hermannleboss:
# Description

## Initial Issue Description

Fixes #1854

`app/src/composables/useSiwe.ts` was using `useFetch` from `@vueuse/core` for the SIWE login flow (nonce + auth), which was inconsistent with the rest of `app/` (TanStack Query + Axios via `apiClient` and the `createQueryHook` / `createMutationHook` factories).

## PR Summary Or Solution description

Implements **Option B** from the issue:

- **`apiClient` retry interceptor** ([axios.ts](../blob/claude/laughing-babbage-2d19b2/app/src/lib/axios.ts)) — response interceptor with 3 retries / 500 ms backoff on `408/425/429/500/502/503/504` and network errors. Gives every consumer cold-start retries "for free" and lets us drop hand-rolled loops at the call site.
- **`useSiweAuthMutation`** ([auth.queries.ts](../blob/claude/laughing-babbage-2d19b2/app/src/queries/auth.queries.ts)) — `POST /auth/siwe` via `createMutationHook`, returns `{ accessToken }`.
- **`getUserNonce(address)`** ([user.api.ts](../blob/claude/laughing-babbage-2d19b2/app/src/api/user.api.ts)) — thin axios wrapper for the imperative nonce fetch. The SIWE flow is sequential, so a reactive query with `enabled: false` would be artificial here.
- **`useSiwe.ts` migration** — drops `useFetch` and `BACKEND_URL`, calls `getUserNonce` then `siweAuthAsync({ body: { message, signature } })`, then refetches the user via the existing TanStack hook.

Atomic commits:
1. `feat: ✨ retry transient failures in apiClient`
2. `feat: ✨ add SIWE auth mutation and nonce API wrapper`
3. `refactor: ♻️ migrate useSiwe to TanStack Query + Axios`
4. `test: ✅ add unit tests for useSiweAuthMutation`

## Acceptance criteria (from #1854)

- [x] No `useFetch` import remaining in `app/src/composables/useSiwe.ts`.
- [x] SIWE POST goes through a `createMutationHook`-based mutation in `auth.queries.ts`.
- [x] Nonce fetch goes through `apiClient` (via `getUserNonce` → `apiClient.get`).
- [x] First-call cold-start failures retry automatically (axios interceptor; no hand-rolled loop in `useSiwe`).
- [x] Existing SIWE specs still pass; new tests added for `useSiweAuthMutation` (happy path + error propagation).

[...]

PR #1971, opened 2026-05-18
feat: ✨ e2e login flow with a wallet mock connector

### LINKED ISSUE #1914: feat: ✨ Playwright e2e tests with wagmi mock connector (Hardhat-backed)
## Goal

Enable end-to-end testing of wallet-dependent flows (SIWE login, contract calls, Safe operations) without dealing with the MetaMask browser extension. Tests run against a local Hardhat node with a wagmi `mock` connector swapped in via env flag, so Playwright drives the UI exactly like a real user but signatures and transactions are handled by an in-process test wallet.

## Why this approach

- **MetaMask e2e (Synpress) is flaky in CI**: extension UI selectors break on every MetaMask update, popup orchestration is slow (~30-60s per scenario), and the niche tooling is maintained by a small community.
- **Mock connector covers ~95% of real bugs**: routing, composables state, contract reads/writes, SIWE flow, event listeners, error handling — all exercised against a real chain.
- **Stay on Hardhat (don't switch to Anvil)**: `contract/` already has working `deploy.sh`, Ignition modules, OZ upgrades plugin, and chain `31337` is already wired in `app/src/wagmi.config.ts`. Anvil's faster startup is irrelevant for a Playwright suite.

### 1. wagmi config — env-driven connector swap

`app/src/wagmi.config.ts`: introduce a `VITE_E2E` flag. When true, expose a single `mock` connector seeded with Hardhat's pre-funded account #0. When false (prod/dev), behave exactly as today.

```ts
const isE2E = import.meta.env.VITE_E2E === 'true'
connectors: isE2E
  ? [mock({ accounts: [<hardhat-account-0-address>] })]
  : [injected()]
```

### 2. `useWalletChecks` — read the configured connector

`app/src/composables/useWalletChecks.ts:25` currently hardcodes `injected()`. Switch to `config.connectors[0]` so the env flag actually takes effect at the call site.

### 3. Playwright setup

- Add `@playwright/test` as a devDep in `app/`.
- `app/playwright.config.ts` with a `webServer` block that runs `VITE_E2E=true npm run dev` and waits for the port.
- `app/e2e/` directory for specs. Start with one smoke test: connect → SIWE → land on `/teams`.

### 4. Test fixtures for the chain

A `beforeAll` (or a Playwright global setup) that:
1. Starts Hardhat node (`cd contract && npm run node`) — or assumes one is running locally.
2. Runs `npm run deploy` to deploy the contract suite.
3. Exposes deployed addresses to the app via env or a generated JSON read at boot.

Stretch: use Hardhat's `evm_snapshot` / `evm_revert` between tests for isolation.

### 5. CI

GitHub Actions job that boots Hardhat + the Vite dev server + Playwright in one workflow. Cache `node_modules` and Playwright browsers.

## Out of scope

- True MetaMask extension testing (Synpress) — can be added later as a thin smoke layer on top if needed.
- Replacing Vitest unit/integration tests — Playwright is additive, not a replacement.
- Migrating from Hardhat to Anvil.

## Acceptance criteria

- [ ] `VITE_E2E=true npm run dev` boots the app with the mock connector active and no MetaMask required.
- [ ] At least one passing Playwright spec that covers: visit landing → click "Connect" → SIWE signature succeeds → user is authenticated and lands on the post-login route.
- [ ] CI job runs the e2e suite on PRs touching `app/` or `contract/`.
- [ ] README / `AGENTS.md` updated with the e2e workflow (how to run locally, where to add new specs).

## References

- Discussion in chat: trade-off Synpress vs mock connector, Hardhat vs Anvil.
- Current config: [app/src/wagmi.config.ts](https://github.com/globe-and-citizen/cnc-portal/blob/develop/app/src/wagmi.config.ts), [app/src/composables/useWalletChecks.ts](https://github.com/globe-and-citizen/cnc-portal/blob/develop/app/src/composables/useWalletChecks.ts).

hermannleboss:
Closes #1914.

## What

Adds a Playwright e2e test for the SIWE login flow, driven by an in-browser
wagmi **mock connector** instead of the MetaMask extension.

When the app runs with `VITE_E2E=true`, `wagmi.config.ts` registers
`e2eMockConnector` (`app/src/e2e/mockConnector.ts`) — a connector that wraps a
viem local account (Hardhat test account #0) and handles `connect`,
`switchChain` and message signing **in-page**. Signing is done locally with
the test key, so the login flow needs no running chain and the test is
hermetic. Dev and production builds are unaffected: they register no
connector here and keep using `injected()` (the mock connector and its
`viem/accounts` import are tree-shaken out of the production build).

## Changes

- **`feat`** — `e2eMockConnector` + `VITE_E2E` wiring in `wagmi.config.ts`;
  `viem/accounts` added to `optimizeDeps.include` so Vite never re-optimizes
  mid-session (a cold-start re-optimization otherwise broke the run with an
  "Outdated Optimize Dep" reload).
- **`fix`** — `performChecks` fired `connect`/`switchChain` as
  fire-and-forget mutations and read connection state synchronously, so the
  wallet check always failed on the first attempt (login only worked on a
  second click). It now awaits the mutations. It also picks the
  pre-registered connector when present, falling back to `injected()`.
- **`test`** — `login.spec.ts` rewritten as a plain Playwright flow (stub
  backend → click sign-in → assert redirect to `/teams`); dev server started
  with `VITE_E2E=true`.
- **`docs`** — e2e README rewritten around the mock-connector workflow;
  `AGENTS.md` command reference updated.

## Verification

- `npm run test:e2e` — login spec passes from a cold dev server (~10s).
- `npm run lint`, `npm run format-check` — clean.
- `npm run test:unit -- --run` — 1682 tests pass.
- `npm run type-check` — clean for all files in this PR. Note: 4 pre-existing
  `noImplicitAny` errors in `CreateElectionForm.vue` / `CreateProposalForm.vue`
  exist on `develop` and are unrelated to this change (tracked separately).

## Notes / follow-ups

- This PR keeps `@synthetixio/synpress` as a (now-unused) dependency.
  Removing it cleanly requires regenerating `package-lock.json`, which the
  available npm version rewrites wholesale (~18k lines) — out of scope here.
  A follow-up should drop the dependency, its `test:build:cache*` scripts and
  the `test/wallet-setup/` scaffolding.


app/src/composables/useSiwe.ts
```diff
@@ -2,115 +2,157 @@ import router from '@/router'
...
+      // `address` and `chainId` are non-nullable.
+      const { address, chainId } = getConnection(config)
+      if (!address || chainId === undefined) {
+        throw new WalletConnectRejectedError(new Error('No active connection after switch'))
       }
 
+      const { nonce } = await withRetry(
+        () => getUserNonce(address),
+        NONCE_RETRIES,
+        NONCE_RETRY_DELAY_MS
```
hermannleboss: [PR #1856](https://github.com/globe-and-citizen/cnc-portal/pull/1856) added a response retry interceptor on `apiClient` (see [`app/src/lib/axios.ts` L37-L56](https://github.com/globe-and-citizen/cnc-portal/blob/27a6dd98a6f9e02ec6ec0e0cb7b080293b8c56c5/app/src/lib/axios.ts#L36-L57)) precisely so callers could drop hand-rolled retry loops. `getUserNonce` goes through `apiClient.get`, so this `withRetry(…, 3, 3000)` is now stacking on top of the interceptor:

- a flaky nonce fetch can retry **up to 9 times total** (3 by the interceptor with 500ms backoff, then 3 more by `withRetry` with 3s backoff)
- a non-retriable failure (400/401/404) that the interceptor correctly skips is still retried 3 times by `withRetry`, costing 6 seconds of dead air before the user sees the toast

Question: does the SIWE nonce fetch actually need a retry policy different from every other API call — and if not, can the interceptor own this and the `withRetry` wrapper go away?

hermannleboss (commented):
### Code review

Found 2 issues:

1. **Hand-rolled `withRetry` stacks on top of the `apiClient` retry interceptor.** See inline comment on `useSiwe.ts`. The interceptor was added in [PR #1856](https://github.com/globe-and-citizen/cnc-portal/pull/1856) precisely so callers could drop these wrappers — `getUserNonce` goes through `apiClient.get`, so a flaky nonce fetch can now retry up to **9× total** (3 from the interceptor with 500ms backoff, then 3 from `withRetry` with 3s backoff), and any non-retriable error (400/401/404) that the interceptor correctly skips is still retried 3× with 6s of dead air.

2. **Dead mock plumbing for the removed `useWalletChecks` composable.** The PR deletes `app/src/composables/useWalletChecks.ts` and removes its export from `app/src/composables/index.ts`, but `mockUseWalletChecks` and the `vi.mock('@/composables', { useWalletChecks: … })` shim still survive in the test setup. AGENTS.md, [“Surface drift proactively”](https://github.com/globe-and-citizen/cnc-portal/blob/27a6dd98a6f9e02ec6ec0e0cb7b080293b8c56c5/AGENTS.md), explicitly calls out this kind of dead-reference drift (the example given is “dead useToastStore references”).

  - https://github.com/globe-and-citizen/cnc-portal/blob/27a6dd98a6f9e02ec6ec0e0cb7b080293b8c56c5/app/src/tests/mocks/composables.mock.ts#L158-L170
  - https://github.com/globe-and-citizen/cnc-portal/blob/27a6dd98a6f9e02ec6ec0e0cb7b080293b8c56c5/app/src/tests/setup/composables.setup.ts#L273-L283

  Question: any reason to keep the mock around (in case `useWalletChecks` comes back), or can it go in the same PR as the deletion?

app/src/composables/useSiwe.ts
hermannleboss: Addressed in 43cd8d868: dropped `withRetry` + `NONCE_RETRIES` / `NONCE_RETRY_DELAY_MS` / `sleep` helpers from `useSiwe.ts`, and doubled `RETRY_DELAY_MS` from 500ms to 1000ms in `lib/axios.ts` so the interceptor's budget (3 × 1000ms ≈ 3s) covers a typical Railway cold start uniformly for all callers.

hermannleboss (commented):
LGTM

PR #2013, opened 2026-05-25
refactor: ♻️ mutation composable conventions (toast/invalidation ownership + typed-error contract)

### LINKED ISSUE #1776: Pattern guidelines: pure async function + TanStack mutation composable
## Context

We've adopted a pattern across several contract-interaction composables: **extract the async logic as a pure function, then wrap it in a TanStack `useMutation` composable**. Current examples:

- `app/src/composables/contracts/useOfficerDeployment.ts` — `deployOfficer()` + `useDeployOfficer()`
- `app/src/composables/investor/useShareholderMigration.ts` — `migrateShareholders()` + `useMigrateShareholders()` + `InconsistentSupplyError`
- `app/src/composables/contracts/useOfficerRedeploy.ts` — workflow composable that orchestrates three mutations

The pattern works well and should be the default for new write-side composables. But there are **subtle pitfalls** that will cause bugs as the codebase grows. This issue documents them as a checklist so we can audit existing code, write docs, and enforce conventions in review.

---

## Why this pattern (what we gain)

- **Pure function is trivially testable** — no Vue runtime, no TanStack, just input/output. Usable from scripts, workers, E2E tests without mounting components.
- **TanStack provides standard state for free** — `isPending`, `error`, `data`, `reset`, `isSuccess`, `isError`, `variables`, plus DevTools integration and automatic cleanup on unmount.
- **Typed errors via `class extends Error`** (e.g. `InconsistentSupplyError`) let consumers narrow with `instanceof` and react precisely, instead of switching on magic status strings.
- **Composition** — orchestrators (e.g. `useOfficerRedeploy`) combine multiple mutations with `OR` on `isPending`, without a custom state machine.

---

### 1. Double-toast risk when composables stack

**Problem.** If `useDeployOfficer.onSuccess` emits a toast (`"Officer contract deployed successfully"`) AND an orchestrator or a component that awaits `mutateAsync()` also emits one for its own flow step, the user sees two toasts for what is semantically a single action.

**Concrete example in the codebase.**
- `useDeployOfficer` (`useOfficerDeployment.ts` onSuccess) toasts `"Officer contract deployed successfully"`.
- `useOfficerRedeploy.redeploy()` awaits that same `mutateAsync`, then later toasts `"Officer redeployed and contracts synced"` when the whole flow finishes.
- Today the messages differ, so it's not wrong — but nothing prevents a future maintainer from adding an identical toast in both places.

**Recommendation.**
- **Rule of thumb:** *wrappers toast low-level atomic outcomes (RPC error, user rejection, basic success). Orchestrators / components toast business-flow outcomes.*
- When composing, silence the wrapper-level success toast if the orchestrator will issue its own. Either:
  - Accept `options` on the wrapper (`useDeployOfficer(options?)`) and let the caller override `onSuccess`/`onError`.
  - Or move all toasts out of the wrappers entirely (strict version — see §5).
- Add a checklist item in PR review: *"Does this call path emit more than one toast per logical action?"*

---

### 2. Redundant cache invalidation

**Problem.** Each mutation in a multi-step flow runs its own `onSuccess` invalidation. When orchestrators chain mutations, the same query keys are invalidated multiple times in a row.

**Concrete example in the codebase.**
During `useOfficerRedeploy.redeploy()`:
1. `useDeployOfficer.onSuccess` invalidates `['team', id]`, `['teams']`, `['contracts']`.
2. `useCreateOfficerMutation.onSuccess` (via `createMutationHook`'s `invalidateKeys: [contractKeys.all, teamKeys.all]`) invalidates roughly the same keys.
3. The orchestrator calls `useInvalidateOfficerQueries()` at the end — invalidates them again.

So three back-to-back refetches where one would do. Not wrong, just wasteful, and on slow connections the UI flickers.

**Recommendation.**
- **Rule of thumb:** *wrappers invalidate ONLY the query keys strictly scoped to their own effect. Orchestrators own the "flush everything at the end" invalidation.*
- Audit each wrapper: does it invalidate keys that a downstream wrapper will also touch? Tighten it.
- `useDeployOfficer.onSuccess` could drop `['team', id]` invalidation — the `registerOfficer` mutation that follows will hit it anyway. Deploy alone doesn't change backend team data.
- Consider adding an `options.skipInvalidation?: boolean` on wrappers used inside orchestrators.

---

### 3. Workflow state still needs manual refs

**Problem.** TanStack gives per-call state (`isPending`, `data`, `error` for *one* mutation). It doesn't cover state that **spans multiple mutations** in a workflow.

**Concrete example in the codebase.**
`useOfficerRedeploy.pendingMigration` is a local `ref<{...} | null>` — it holds addresses that must survive across the register mutation's completion and the migrate mutation's start/failure. No single TanStack mutation can express "we are past step 2 and stalled on step 3 awaiting user decision".

**Recommendation.**
- Accept that orchestrators will always have some custom refs. That's fine — it's what workflow composables exist for.
- **Convention:** derive as much as possible from the mutations themselves:
  - `isRunning = OR(mutationA.isPending, mutationB.isPending, ...)`.
  - `failedAt = computed(() => firstMutationWithError)`.
- Keep manual refs **only** for state that is genuinely multi-call (addresses to retry with, pending-confirmation flags). Don't re-invent `isPending` with a local ref.

---

### 4. Coupling to the `mutateAsync` / `isPending` API surface

**Problem.** Components and orchestrators read `mutation.isPending.value`, `mutation.mutateAsync`, `mutation.error.value`. If we ever migrate off `@tanstack/vue-query` (to Pinia Colada, Nuxt's built-in data layer, a custom layer), every call site has to be rewritten.

**Recommendation.**
- Low priority — TanStack is battle-tested and we're unlikely to migrate soon.
- But if we want extra insurance, consumers could destructure to our own names: `{ mutateAsync: deploy, isPending: isDeploying } = useDeployOfficer()`. This is already done in some places (e.g. `DeployContractSection.vue`) — let's standardize it as a habit.
- An even safer version: the wrapper could return a narrower custom shape (`{ run, isRunning, error, data }`) that hides TanStack. Probably overkill for us, but worth naming as an option.

---

### 5. Two valid places for side effects — pick one per wrapper

**Problem.** Side effects (toasts, logs, invalidation) can live either:
- **A.** In the wrapper composable's `onSuccess` / `onError`.
- **B.** At the call site, after `await mutateAsync()`.

Both patterns work. But mixing them within a single flow produces duplicate effects (see §1) or lost effects (when a caller assumes the wrapper handles a toast, and it doesn't).

**Concrete example in the codebase.**
- `useDeployOfficer` uses pattern A (toast inside `onSuccess`).
- `DeployContractSection.vue` uses pattern B for its own "deployed and synced successfully" toast after the full flow.
- `useCreateOfficerMutation` uses pattern A via `createMutationHook`'s `invalidateKeys`.

Result: a mix of A and B on the same flow. No bug today, but each new dev has to infer the convention.

**Recommendation.**
- Pick a **default** per-wrapper stance:
  - **Wrapper owns basic side effects** (low-level toast, narrow invalidation). Callers extend, don't duplicate.
  - OR, **stricter version**: wrappers are pure plumbing, all side effects at the call site. More verbose but orthogonal.
- Document the choice at the top of each file:
  ```ts
  // Side effects: this wrapper toasts on deploy success/failure and
  // invalidates ['contracts']. Callers must NOT re-toast success.
  ```
- Enforce in review.

---

### 6. Error typing is fragile if we mix Error subclasses and strings

**Problem.** We introduced `InconsistentSupplyError` and rely on `error instanceof InconsistentSupplyError` in the UI. This works, but:
- If the error is caught, re-wrapped, or serialized (e.g. through a Sentry transport, a `parseError` helper), the `instanceof` relationship can be lost.
- If a second wrapper rewraps the error as a plain `Error` with a concatenated message, consumers silently degrade to generic handling.

**Recommendation.**
- Never catch-and-rethrow a typed error as a plain `new Error(msg)`. Preserve the original: `throw new WrapperError(msg, { cause: originalTypedError })`.
- Consumers checking for specific types should walk the `cause` chain, not just the top-level error.
- Add a small test covering "typed error survives the full mutation pipeline" for every typed error we introduce.

---

### 7. Return-type discriminated unions vs. thrown errors — pick intentionally

**Problem.** `migrateShareholders` returns `{ kind: 'done' | 'noop-empty' | 'noop-already-migrated' }` for success outcomes but **throws** `InconsistentSupplyError` for the blocked case. This is a reasonable choice (blocked is a recoverable error, while noop isn't), but it's a subtle API.

Consumers have to remember to:
- `switch(result.kind)` on success outcomes
- AND `try/catch` (or inspect `mutation.error`) for the blocked / RPC errors

If a dev adds a new outcome and isn't sure which side (return vs throw) it belongs on, the API drifts.

**Recommendation.**
- Document the rule in the function's JSDoc:
  > *Throws for conditions that require user intervention or are unrecoverable in this run (inconsistent state, RPC failure, user rejection). Returns a `kind` for outcomes that completed without needing a transaction (noop-empty, noop-already-migrated).*
- Review each new outcome against this rule before adding.

---

### 8. Reactive dependencies inside pure functions are invisible

**Problem.** The pure function takes plain addresses / bigints as arguments, not refs. That's the right design. But the *caller* has to remember to unwrap refs at the call site — easy to forget with TypeScript helping less than you'd expect (`unref` returns `T | T`, so bugs pass tsc).

**Concrete example.**
In `ShareholderMigrationBanner.vue`:
```ts
await migrate.mutateAsync({
  previousOfficerAddress: previousOfficerAddress.value,
  newInvestorAddress: currentInvestorAddress.value as Address
})
```
If someone later removes `.value`, TypeScript still accepts it (wide `any`-like inference in some places), and the pure function receives a `Ref<Address>` treated as `Address`.

**Recommendation.**
- Keep pure function signatures strict and narrow (`Address`, not `Address | Ref<Address>`).
- Add ESLint rule or a test that forbids `Ref` objects crossing into mutation arguments.

---

## Acceptance criteria

- [ ] Add this file: `app/src/composables/CONVENTIONS.md` — concise restatement of the rules in §1 (toast ownership), §2 (invalidation ownership), §5 (side-effect placement), §7 (throw vs return).
- [ ] Audit existing wrappers and tighten invalidation scope where redundant (§2): list the ones that currently over-invalidate and either drop the extra keys or document why they're needed.
- [ ] Each wrapper file gets a top-of-file comment declaring its side-effect contract (§5 recommendation).
- [ ] Add one targeted test per typed error class ensuring the `instanceof` check survives the TanStack pipeline (§6).
- [ ] PR review checklist updated with: *"Are there duplicate toasts / invalidations? Does side-effect placement follow the wrapper's declared contract?"*
- [ ] Optional: introduce `options?: { onSuccess?, onError?, skipInvalidation? }` on shared wrappers used by orchestrators (§1, §2).

## Out of scope

- Migrating off `@tanstack/vue-query` (§4 is documentation only).
- Rewriting `useOfficerRedeploy` as a state machine library — the manual ref approach is acceptable (§3).
- Read-side composables (this issue is scoped to mutation wrappers).

hermannleboss:
## Summary

Operationalizes the mutation-pattern guidelines from #1776 against the canonical wrappers and the redeploy orchestrator.

- **`app/src/composables/CONVENTIONS.md`** — concise review checklist covering toast ownership (§1), invalidation ownership (§2), workflow state (§3), side-effect contract header (§4), typed errors (§5), throw vs return (§6), argument plumbing (§7).
- **`useDeployOfficer` / `useInvalidateOfficerQueries`** — drops the dead `['team', id]` query key (singular `team`, never matched any registered query), switches to `teamKeys.all` + `contractKeys.all` factories, adds `{ silent, skipInvalidation }` options. Side-effect contract header added.
- **`useMigrateShareholders`** — adds `{ silent }` option, documents that errors are never toasted by the wrapper (callers render via UAlert / `instanceof InconsistentSupplyError`). Side-effect contract header added.
- **`useOfficerRedeploy`** — now silences both wrappers and skips their per-step invalidation, so the user sees exactly one flow-level toast and one cache flush per redeploy. Adds a flow-level toast on retry success (the wrapper toast it used to lean on is silenced).
- **`useShareholderMigration.pipeline.spec.ts`** — new spec that bypasses the global `useMutation` mock with `vi.importActual` and asserts `mutation.error.value instanceof InconsistentSupplyError` after a real TanStack pipeline. Failure means a future refactor silently degraded the typed-error guarantee.
- **`composables.setup.ts`** — global `team.queries` / `contract.queries` mocks now expose the real `teamKeys` / `contractKeys` shapes so wrappers that invalidate via key factories stay testable.

Closes #1776.

## Test plan

- [x] `npm run type-check` — no new errors (`useToast` pre-existing 65 → 65)
- [x] `npx vitest run` — 208 files / 1702 tests pass
- [x] `npx eslint <touched files>` — clean
- [x] `npx prettier --check <touched files>` — clean
- [ ] Manual smoke: trigger Officer redeploy with a previous Officer present, confirm only one success toast and one refetch wave on the network panel
- [ ] Manual smoke: `ShareholderMigrationBanner` retry path still toasts on success


hermannleboss (commented):
LGTM

Issue #1845, opened 2026-04-29
refactor: ♻️ split composables.setup.ts and stop globally mocking TanStack Query / wagmi


hermannleboss:
Parent: #1837

`composables.setup.ts` is 319 lines covering all domains. `useQuery`/`useMutation`/`useQueryClient` are mocked globally → impossible to exercise real TanStack flows without `vi.unmock`. wagmi/contracts are also mocked at module level, forcing 7-10 specs to re-mock them locally.

Depends on: renderWithProviders + factories landing first.

## Tasks

- Split `composables.setup.ts` by domain (`auth.setup.ts`, `safe.setup.ts`, `balance.setup.ts`, …).
- Remove the global mocks of `@tanstack/vue-query`, `@wagmi/vue`, `@wagmi/core`, contract modules.
- Provide opt-in helpers: `mockQuery({ data, isPending, error })`, `mockMutation(...)`, `mockWagmiAccount(...)`.
- Migrate specs that previously relied on global mocks to call the helpers explicitly.
- Update setupFiles list in `vitest.config.ts`.

## Acceptance

- `composables.setup.ts` deleted or reduced to a thin re-export.
- TanStack Query / wagmi no longer mocked globally.
- All specs green.
- Coverage of real composables increases (measurable).

hermannleboss:
Closing as won't-do. Challenging the premise: removing the global `@tanstack/vue-query` mock breaks 24 spec files / ~209 tests with `No 'queryClient' found in Vue context`. Categorising the fallout: ~20 of those files (component/composable specs) would only need mechanical `QueryClient` re-wiring at mount — no extra logic exercised, since `@/queries/*` stays mocked next to it. Only ~3-4 files (the `@/queries/*` and store specs) genuinely benefit, and those already `vi.unmock` locally today.

Conclusion: the real value isn't "delete the global mock" but "test the `@/queries/*` layer for real", which is a much smaller, targeted effort and doesn't require this umbrella task. Keeping the global mock; closing PR #2026 too.

Issue #1843, opened 2026-04-29
test: ✅ create renderWithProviders test helper


hermannleboss:
Parent: #1837

136 specs duplicate `global: { plugins: [...] }`. `mountWithProviders` exists in `nuxt-ui.setup.ts` but is barely used and not generic enough.

## Tasks

- Create `app/src/tests/helpers/renderWithProviders.ts`:
  - Mounts with `createTestingPinia({ createSpy: vi.fn })` by default, overridable.
  - Installs `TooltipProvider` and Nuxt UI stubs.
  - Provides a real or mocked `QueryClient` (opt-in).
  - Accepts a parameterizable route stub instead of the global hard-coded `{ id: '1' }`.
- Document usage in `app/src/tests/README.md` with a before/after example.
- Migrate ~10 representative specs as proof of value.

## Acceptance

- Helper exported and typed.
- ≥ 10 specs migrated, net LOC reduced.
- Documented as the default mounting path.

PR #2026, opened 2026-05-29
refactor: ♻️ extract TanStack Query mock into opt-in setup (#1845)

hermannleboss:
## Summary

First, non-breaking slice of #1845 (stop globally mocking TanStack Query / wagmi).

- Move the global `@tanstack/vue-query` mock out of the `composables.setup.ts` monolith into a dedicated `tanstack.setup.ts`, continuing the per-domain split of the test setup.
- Add an opt-in `withQueryClient` / `createTestQueryClient` helper (`app/src/tests/helpers/queryClient.ts`) so specs can mount with a real `QueryClient` and exercise real query/mutation flows.
- Document the opt-in path and the migration plan in `app/src/tests/README.md`.

Behaviour is unchanged — the mock is still global. This is the groundwork before the breaking step.

## Why scoped this way

Measured the blast radius of simply deleting the global mock: **23 spec files / 209 tests fail** with `No 'queryClient' found in Vue context` (components call `useQueryClient()` directly). Migrating those in one PR would be large and depends on `renderWithProviders` (#1843). So this PR ships the foundation green; the deletion + per-spec migration follows.

## Test plan

- [x] `npm run type-check` clean
- [x] `npm run lint` clean on changed files
- [x] Full unit suite green (213 files / 1748 tests)

Relates to #1845

### DIFF
--- a/app/src/tests/README.md
+++ b/app/src/tests/README.md
@@ -23,6 +23,27 @@ ESLint enforces this via `no-restricted-syntax`: any `vi.mock('<globally-mocked-
 
 If you need to mock a module that **isn't** yet globally mocked but you're reaching for it across several specs, add it to `src/tests/setup/` and re-export the override hook from `src/tests/mocks/index.ts` rather than re-mocking it in each spec.
 
+## Opt-in: exercising real TanStack Query flows
+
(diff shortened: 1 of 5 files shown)


hermannleboss:
Closing without merging. After measuring the impact (see #1845), we've decided to keep the global TanStack Query mock: removing it breaks 24 spec files / ~209 tests, and ~80% of that work is mechanical re-wiring of a QueryClient at mount time with no real gain in test quality (the `@/queries/*` layer stays mocked alongside). Not worth the churn.

PR #2033, opened 2026-05-30
fix: 🐛 remove hardhat/console.sol imports from production contracts

### LINKED ISSUE #1992: fix: 🐛 remove hardhat/console.sol imports from production contracts
`hardhat/console.sol` is a dev-only import shipped to production. Bloats bytecode, leaks dev intent, and on some networks can be a footgun.

## Files
- `contract/contracts/CashRemunerationEIP712.sol:16`
- `contract/contracts/Voting/Voting.sol:5`

## Scope
- Remove the imports and any `console.log` call sites.
- Add a `solhint` rule (or grep step in CI) to fail any new import of `hardhat/console.sol` in `contracts/` (excluding `contracts/mocks/` and `contracts/test/`).

## Acceptance
- [ ] No `hardhat/console.sol` import outside dev/mock contracts.
- [ ] CI guard added.
- [ ] Bytecode size for affected contracts decreases (sanity check).

hermannleboss:
## Summary
- Remove the dev-only `hardhat/console.sol` import from `CashRemunerationEIP712.sol` and `Voting.sol` — it was shipped to production, bloating bytecode and leaking dev intent.
- Add a CI guard in `contract-code-quality.yml` that fails any import of `hardhat/console.sol` under `contracts/`, excluding `contracts/mocks/` and `contracts/test/`.

## Why
`hardhat/console.sol` is a development helper that should never reach production contracts. The guard prevents regressions.

## Notes for reviewer
- There were **no `console.log` call sites** — only the imports. As a result the deployed runtime bytecode is effectively unchanged (the import alone adds no opcodes without a `console.log` call), so the issue's "bytecode decreases" sanity check doesn't apply here. The real wins are removing the dev footgun and the CI guard.
- Verified locally: `npm run compile` (84 files), `npm run lint:sol`, and `npm run format-check:sol` all pass; the guard greps clean.
- Note: PR #2002 also targets this issue — one should be closed to avoid duplication.

Closes #1992

### DIFF
--- a/.github/workflows/contract-code-quality.yml
+++ b/.github/workflows/contract-code-quality.yml
@@ -34,6 +34,19 @@ jobs:
         run: npm run format-check
         working-directory: ./contract
 
+      - name: 🚫 Guard against hardhat/console.sol in production contracts
+        run: |
+          matches=$(grep -rn "hardhat/console.sol" contracts \
+            --include="*.sol" \
+            --exclude-dir=mocks \
+            --exclude-dir=test || true)
+          if [ -n "$matches" ]; then
+            echo "::error::hardhat/console.sol must not be imported outside contracts/mocks and contracts/test"
+            echo "$matches"
+            exit 1
+          fi
+        working-directory: ./contract
+
       - name: </> Run Solidity Linter
         run: npm run lint:sol
         working-directory: ./contract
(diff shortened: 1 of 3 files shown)


hermannleboss (commented):
LGTM

PR #2064, opened 2026-06-04
feat: ✨ pagination prototype in app (WeeklyClaim) #2028

Georginio-prod:
# Description

## Initial Issue Description

App-side of the pagination prototype, now rolled out across **all** app transaction tables. The dashboard side ([#2029](https://github.com/globe-and-citizen/cnc-portal/issues/2029)) is already merged; this brings the same behaviour to the `app/` SPA.

Fixes #2028 (parent: #2027)

## PR Summary Or Solution description

Introduces a single route-bound pagination primitive and applies it everywhere a table paginates, so the whole app behaves consistently.

**New `usePagination` composable** (`app/src/composables/usePagination.ts`)
- `page` + `pageSize` read from / written to the **route query** via `router.replace` → shareable links, reload-persistent, no history spam.
- **Resize anchoring**: when the page size changes the item at `rangeStart` stays in view — `newPage = ceil(rangeStart / newPageSize)`.
- Default-valued params dropped from the URL; out-of-range sizes fall back to default; optional `key` prefix lets several tables share one route.

**New `TablePagination` footer** (`app/src/components/TablePagination.vue`)
- `Showing {{ rangeStart }}–{{ rangeEnd }} of {{ total }} {{ noun }}` summary + page-size selector + Nuxt UI `UPagination` (next/prev, first/last, jump-to-page, disabled boundary states). Renders nothing when `total === 0`.

**Rolled out to every paginated table** (replacing the ad-hoc `TransactionTableFooter`):
- `useTransactionTable` now sources `page`/`pageSize` from `usePagination` (route-bound + anchoring) via an optional `key`; dropped the old `pageSize → reset` watch.
- `SafeTransactions` (own refs) migrated to `usePagination` (key `safeTx`).
- `Bank` / `Expense` / `CashRemuneration` / `Investor` transactions → `TablePagination` + unique keys.
- `WeeklyClaim` namespaced with key `weeklyClaim` — it shares the Company Payroll route with `CashRemunerationTransactions`, so bare params would collide.
- Deleted `TransactionTableFooter.vue` and its now-dead `PAGE_SIZE_OPTIONS` export.

**Test harness:** `mockRoute` is now reactive and the mock `router.replace` self-updates the query (query-only), so route-bound pagination is exercised by the existing component/composable specs. Added 14 `usePagination` unit tests; updated the Expense stub to `TablePagination`.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Refactor (unifies pagination across the app)

[...]

PR #2058, opened 2026-06-03
feat: ✨ pagination prototype in dashboard (#2029)

Georginio-prod:
# Description

## Intial Issue Description

Prototype the unified pagination behaviour end-to-end on the dashboard accounting tables.

Fixes #2029
Parent: #2027

## PR Summary Or Solution description

The dashboard already had a pagination footer (`Showing X–Y of Z` + page-size selector + `UPagination` controls), but it was missing the two behaviours required by #2029:

1. **Resize anchoring** — changing the page size used to reset to page 1. It now keeps the item currently at `rangeStart` in view: `newPage = ceil(rangeStart / newPageSize)`. Example (total = 100): page 2 / size 10 (range 11–20) → size 5 lands on page 3 (range 11–15); size 20 lands on page 1 (range 1–20).
2. **Route / URL binding** — `page` + `pageSize` are now mirrored to the route query, so a page link is shareable and survives a reload.

To keep the three accounting tables (General Ledger, Positions, Income Statement) consistent — they live on the **same** `accounting.vue` route in different tabs — the logic was extracted into a `usePagination` composable. Each table passes a distinct `key` (`ledger` / `positions` / `income`) so their query params don't collide.

### Implementation notes
- **`dashboard/app/composables/usePagination.ts`** (new): owns `page` + `pageSize` bound to the route query via `router.replace` (no history spam, follows the existing `useWalletAddress` pattern). Implements resize anchoring in the `pageSize` setter, validates the page size against the allowed options, and drops a param from the URL when it equals its default so a pristine table keeps a clean URL.
- **`AccountingPagination.vue`**: now purely presentational — it no longer forces page 1 on a size change; the owner (the composable) decides which page to land on.
- **`AccountingLedger.vue`, `AccountingPositions.vue`, `AccountingIncomeStatement.vue`**: migrated from local `ref(1)/ref(20)` to `usePagination(...)`, reusing the returned `reset()` in the existing filter-change watchers.

This is a prototype to validate the behaviour, not the final shared abstraction — the reusable composable/UI shared with `app/` is extracted in the refactor sub-issue (#2030).

## Type of change

- [x] New feature (non-breaking change which adds functionality)

- **Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (prototype; tests for resize-anchoring + route binding are planned in #2030)
- [ ] Docs have been added / updated

## Reviewer notes

[...]

Issue #2027, opened 2026-05-29
feat: ✨ unified pagination behaviour for app & dashboard


hermannleboss:
## Context

Pagination is currently re-implemented ad hoc per table (e.g. Company Payroll `WeeklyClaim`, see #2018 / PR #2017). We want a **single, shared pagination behaviour** used across both the `app/` SPA and the `dashboard/`.

This is the **parent issue** describing the expected behaviour of the pagination feature. Implementation will be split into sub-issues.

### Range display
- Display `Showing {{ rangeStart }}–{{ rangeEnd }} of {{ total }}`.

### Page size selection
- The user can choose the number of items per page (page-size selector).

### Resize anchoring (keep the current content in view)
When the page size changes, the item currently at `rangeStart` must stay on the page (fully or partially). The new page is computed so it still contains the previous `rangeStart`:

```
newPage = ceil(rangeStart / newPageSize)
```

Examples (`total = 100`):
- Start: page 2, size 10 → range **11–20**
- Switch to size 5 → range **11–15** (page 3)
- Switch to size 20 → range **1–20** (page 1)

In short: the current page content must remain included (partially or fully) after a page-size change.

### Route / URL binding
- Pagination state (page + page size) must be reflected in the route (query params) so that a page link is **shareable** and survives a reload.

### Remaining standard pagination features
- Next / previous, first / last, jump to page
- Disabled states at boundaries
- Sensible default page size + allowed page-size options
- Empty list / single-page handling

## Scope
- Shared behaviour (composable) + UI usable by both `app/` and `dashboard/`.
- Backend already returns `{ data, total }` for paginated endpoints (see PR #2017); follow that contract.

## Out of scope / follow-ups
- Migrating each existing table to the shared pagination → separate sub-issues.

PR #2069, opened 2026-06-04
refactor: ♻️ roll out shared pagination to dashboard teams & customers #2030

### LINKED ISSUE #2030: refactor: ♻️ roll out shared pagination across app & dashboard
Parent: #2027

## Goal
Once both prototypes (`app/` and `dashboard/`) validate the behaviour, **extract the shared pagination logic** and roll it out across the codebase.

## Tasks
- Extract a reusable pagination composable + UI shared by `app/` and `dashboard/` (range display, page-size selector, resize anchoring, route binding, standard controls).
- Replace the ad hoc per-table implementations with the shared one.
- Ensure every paginated table follows the `{ data, total }` backend contract.
- Add/adjust tests for the resize-anchoring rule and route binding.

## Depends on
- App prototype sub-issue
- Dashboard prototype sub-issue

Georginio-prod:
# Description

## Initial Issue Description

Part of #2030 — *roll out the shared pagination across app & dashboard*. The app side landed via #2064; the dashboard accounting tables already use the shared composable (#2058). This finishes the dashboard roll-out by migrating the **last two tables** that still had ad-hoc pagination.

Fixes #2030 (parent: #2027)

`dashboard/app/components/teams/TeamsList.vue` and `dashboard/app/pages/customers.vue` each rolled their own pagination: a bare `UPagination` wired straight to the TanStack table API, **no page-size selector**, a hard-coded page size of 10, and **no route/URL state**.

Both tables now use the shared **`usePagination`** composable + **`AccountingPagination`** footer (the same ones the accounting tables use), so the whole dashboard behaves consistently: route-bound page + size (shareable, reload-safe), a page-size selector, resize anchoring (`newPage = ceil(rangeStart / newPageSize)`), and the `Showing X–Y of Z` summary.

These are **TanStack-managed** tables (they own filtering / sorting / row selection), unlike the manual-slice accounting tables. So rather than tearing out the table model, pagination becomes a **controlled state bridged to `usePagination`** via a writable computed:

```ts
const { page, pageSize } = usePagination(
  () => table.value?.tableApi?.getFilteredRowModel().rows.length ?? 0,
  { key: 'teams' } // 'customers' for the other
)
const pagination = computed({
  get: () => ({ pageIndex: page.value - 1, pageSize: pageSize.value }),
  set: ({ pageIndex, pageSize: size }) => { /* mirror table → composable */ }
})
```

`getPaginationRowModel` and all TanStack features stay intact; the bare `UPagination` footer is replaced by `AccountingPagination` (for `customers`, the "N of M row(s) selected" line is kept above it). Each table gets a unique query-param `key` so they never collide.

## Reviewer notes

- `customers.vue` is Nuxt UI dashboard-template boilerplate (mock `/api/customers`); migrated for consistency so no table is left on the old pattern.
- **"Shared" = same composable per project, not one cross-project module.** `app/` (Vue + `vue-router`) and `dashboard/` (Nuxt auto-imports) each keep their own identical `usePagination`; a single shared package isn't practical across the framework boundary in this workspace-less monorepo. Worth a follow-up only if the team wants a real shared lib.
- Default page size is now 20 (the composable default), up from the old hard-coded 10 — consistent with every other paginated table.
- No dashboard unit tests added: the `dashboard/` subproject exposes no test runner (only `lint` / `build`). Resize-anchoring + route-binding are covered by the app-side `usePagination` specs.

## Type of change

- [x] Refactor (standardises pagination; non-breaking)

## Code quality gate (dashboard/)

- `npm run lint` (`eslint .`) — clean
- `npm run build` — ✅ build complete (exit 0)
- `npx nuxi typecheck` — no new errors in the touched files (pre-existing dashboard type errors are unrelated)


hermannleboss (commented):
LGTM

PR #2219, opened 2026-06-29
feat: ✨ resolve Ponder addresses from synced deployment artifacts

### LINKED ISSUE #2218: feat: ✨ resolve Ponder contract addresses from synced deployment artifacts
## Problem

The frontend (`app/`, `dashboard/`) auto-resolves contract addresses per `chainId` from `artifacts/deployed_addresses/chain-<id>.json`, which `npm run mc` (in `contract/`) regenerates after a deploy. **Ponder is the exception**: it requires hand-setting `FACTORY_ADDRESS` / `FEE_COLLECTOR_ADDRESS` in `.env.local`. On a local Hardhat node that means manually copying the `OfficerFactoryBeacon` / `FeeCollector` addresses into the env after every redeploy.

## Goal

Make Ponder read addresses from the **same per-network artifact** the frontend uses, synced by `npm run mc`. Env vars stay as optional overrides. Result: `deploy → npm run mc` keeps Ponder in sync with zero manual copy, and addresses are auto-deduced by network.

## Scope (this PR)

- [x] `moveConstants.ts` writes the synced addresses into `ponder/artifacts/deployed_addresses/` too (3rd destination)
- [x] `ponder.config.ts` resolves `Officer#FactoryBeacon` / `FeeCollectorModule#FeeCollector` from `chain-<id>.json` by network, with `FACTORY_ADDRESS` / `FEE_COLLECTOR_ADDRESS` as overrides
- [x] Local `chain-31337.json` is gitignored in `ponder/` (config loads tolerantly when absent → clear "run `npm run mc`" error)
- [x] Docs (`.env.local.example`, `README.md`) updated

## ⚠️ Surfaced inconsistency — Polygon factory needs verification

Ponder hardcodes the Polygon `OfficerFactoryBeacon` as `0x0205fd…`, but the deployment artifact's `Officer#FactoryBeacon` for chain 137 is `0x91EBBe…` — and the **frontend creates teams against the artifact one**. Both addresses are live contracts on Polygon. So today the indexer and the frontend point at different factories.

This PR **keeps Ponder pinned to the historical `0x0205fd…`** (behaviour-preserving) pending verification.

- [ ] Verify on-chain (keyed/archive RPC) which factory actually emitted the `BeaconProxyCreated` / `ContractDeployed` events, then either fix the chain-137 artifact or remove the `FACTORY_PIN` in `ponder.config.ts` so Polygon falls back to the artifact like every other network.

## Follow-up (separate PR)

- [ ] Gitignore + tolerant-load `chain-31337.json` in `app/` and `dashboard/`. Their static JSON imports require the file at build time, so this needs switching to a tolerant import (e.g. `import.meta.glob`) + build/test verification. (The committed `app/` and `dashboard/` `chain-31337.json` are already out of sync with each other — evidence these local artifacts shouldn't be committed.)

hermannleboss:
Closes #2218

## What

Ponder now resolves its contract addresses from the **per-network deployment artifact** (`artifacts/deployed_addresses/chain-<id>.json`) — the same file the frontend consumes, synced by `npm run mc` in `contract/`. No more hand-copying `FACTORY_ADDRESS` / `FEE_COLLECTOR_ADDRESS` into `.env.local` after every local redeploy.

New flow for a local Hardhat node:

```bash
npx hardhat node && npx hardhat run scripts/deploy.ts --network localhost
cd contract && npm run mc      # syncs addresses into app/, dashboard/, ponder/
# ponder/.env.local just needs: NETWORK=hardhat
```

## Changes

- **`contract/scripts/moveConstants.ts`** — adds `ponder/artifacts/deployed_addresses/` as a 3rd sync destination.
- **`ponder/ponder.config.ts`** — reads `Officer#FactoryBeacon` / `FeeCollectorModule#FeeCollector` from `chain-<id>.json` keyed by network. `FACTORY_ADDRESS` / `FEE_COLLECTOR_ADDRESS` env vars become optional **overrides** (`env > artifact`). Loads tolerantly when the file is absent, with a clear "run `npm run mc`" error.
- **`ponder/artifacts/deployed_addresses/`** — seeds the shared networks (137 / 80002 / 11155111); `chain-31337.json` (local Hardhat) is **gitignored** so devs no longer commit per-machine addresses.
- **Docs** — `.env.local.example` and `README.md` updated.

## ⚠️ Polygon factory: behaviour preserved on purpose

The artifact's chain-137 `Officer#FactoryBeacon` (`0x91EBBe…`, what the frontend uses) differs from the factory Ponder has historically indexed (`0x0205fd…`). Both are live contracts on Polygon. To avoid silently re-pointing the production indexer, this PR **keeps Polygon pinned to `0x0205fd…`** via a `FACTORY_PIN` map.

Reconciling which factory is canonical (and then deleting that one-line pin so Polygon falls back to the artifact like every other network) is tracked in #2218 — it needs an on-chain check with a keyed/archive RPC.

## Out of scope (follow-up)

Gitignoring `chain-31337.json` in `app/` and `dashboard/` is **not** included: both statically `import` that JSON, so removing it would break their Vite/Nuxt build. That needs a tolerant-import refactor (`import.meta.glob`) + build/test verification — tracked in #2218.

## Verification

- `npm run typecheck` ✅
- `npm run format-check` ✅
- `npx eslint ponder.config.ts` ✅
- `NETWORK=hardhat npx ponder codegen` loads the config and resolves addresses ✅


hermannleboss (commented):
LGTM

PR #2226, opened 2026-06-30
refactor: ♻️ drive contract table with shared ownable+pausable ABI

### LINKED ISSUE #2225: refactor: ♻️ drive contract table with a shared Ownable+Pausable ABI
## Problem

The contract-management table (`MainContractTable.vue`) only needs each contract’s `owner()` and `paused()` to render, yet `getTeamContracts` in `contractManagementUtil.ts` mapped every contract `type` to its **full ABI** through a `switch`, importing all 7 contract ABIs.

Worse, the `default` branch returned `null`, so any type not explicitly listed — `BoardOfDirectors`, `Campaign`, `VestingV1`, `SafeDepositRouter`, `Safe`, `FixedReturn` — was silently dropped from the table. The table could not discover all of a team’s contracts.

## Desired behaviour

- Use a single minimal **Ownable + Pausable** ABI for every contract instead of importing each contract’s full ABI.
- Discover **all** team contracts in the table, regardless of type.
- Keep `transferOwnership` / `pause` / `unpause` working in `MainContractActions.vue` (it reads `row.abi`).
- A non-Ownable/Pausable contract (e.g. a `Safe`) must not blank the whole table.

hermannleboss:
## What

Replace the per-type ABI `switch` in `getTeamContracts` with a single minimal `OWNABLE_PAUSABLE_ABI` attached to every team contract.

## Why

`MainContractTable` only needs `owner()` / `paused()`, but the util imported all 7 contract ABIs to read them, and its `default` branch returned `null` — silently dropping every unlisted type (`BoardOfDirectors`, `Campaign`, `VestingV1`, `SafeDepositRouter`, `Safe`, `FixedReturn`) from the table.

## Changes

- **New** `app/src/artifacts/abi/ownable-pausable.ts`: minimal ABI = `Ownable.json` + `Pausable.json` + the public `pause()` / `unpause()` entrypoints (not part of OZ `Pausable`). Drives both the table reads and the `MainContractActions` writes (`transferOwnership` / `pause` / `unpause`) read off `row.abi`.
- **Refactor** `contractManagementUtil.ts`: drop the 7 ABI imports and `contractsWithAbis`; attach the shared ABI to every contract so all types are discovered.
- `owner` / `paused` reads are now per-contract tolerant — a non-Ownable/Pausable contract (e.g. a `Safe`) yields `null` instead of rejecting the whole batch and blanking the table.

## Verification

- `build-only` ✅ · `type-check` ✅ · `ContractManagementView` unit tests ✅ (106/106) · prettier ✅

## Note

Non-Ownable contracts (e.g. `Safe`) now appear with `owner: null` / `paused: null`; their action buttons stay disabled (`row.owner !== userDataStore.address`). If we’d rather keep them hidden, we can filter `owner === null`.

Closes #2225

### DIFF
--- a/app/src/artifacts/abi/ownable-pausable.ts
+++ b/app/src/artifacts/abi/ownable-pausable.ts
@@ -1,0 +1,21 @@
+import type { Abi } from 'viem'
+import Ownable from './json/Ownable.json'
+import Pausable from './json/Pausable.json'
+
+// OpenZeppelin's `Pausable` only exposes the `paused()` view; the public
+// `pause()` / `unpause()` entrypoints are declared by the concrete contracts,
+// so they are not part of `Pausable.json`. Add them here so the shared
+// management ABI can drive both the status read and the pause/unpause writes.
+const PAUSE_CONTROLS = [
+  { type: 'function', name: 'pause', inputs: [], outputs: [], stateMutability: 'nonpayable' },
+  { type: 'function', name: 'unpause', inputs: [], outputs: [], stateMutability: 'nonpayable' }
+] as const
+
+/**
+ * Minimal ABI shared by every team contract on the contract-management table.
+ *
+ * Combines Ownable (`owner`, `transferOwnership`, `renounceOwnership`) and
+ * Pausable (`paused`, `pause`, `unpause`). It lets the table read and manage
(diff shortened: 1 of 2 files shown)


hermannleboss (commented):
LGTM

PR #2256, opened 2026-07-06
fix: 🐛 record expenses at withdrawal and keep them visible when the transaction history is behind

### LINKED ISSUE #2257: fix: 🐛 Expenses recorded at approval instead of at withdrawal, and missing from the Accounting page
## Problem

On a team's Accounting page (the general ledger), expenses are not handled correctly, in two separate ways.

First, an expense is written into the books as soon as a spending budget is approved for a member, before any money has actually been spent. This is wrong: an expense should only appear once the money is really taken out. A member can also draw on the same budget several times, and today these repeated withdrawals are not reflected the way they should be. Each withdrawal should show up on its own line, and each one should clearly state how much of the budget is still left (or that the budget has been fully used).

Second, for a team whose transaction history is not up to date, the expenses do not appear on the Accounting page at all. The general ledger shows zero entries under "Expense", even though the spending budgets clearly show that money has already been withdrawn (for example 1000/1000, 134.70/134.70 and 99/200). So the accounting looks completely empty while real money has already left the accounts.

## Expected behaviour

Expenses should be recorded only when the money is actually withdrawn, with one entry per withdrawal, and each entry should say how much of the budget remains. And even when the transaction history is temporarily behind for a team, the Accounting page should still reflect the amounts that have already been spent, so it is never misleadingly empty.

Georginio-prod:
# Description

## description

Expenses are now recorded at the moment the money is actually withdrawn, instead of when the budget is approved. When a member draws on the same budget several times, each withdrawal is recorded on its own line, and every entry states how much of the budget is left (or that it has been fully used).

This PR also fixes the empty-page case. When a team's transaction history is behind, the Accounting page now still reflects the amounts already withdrawn, taken from the spending budgets themselves, so the expenses show up instead of an empty page. As soon as the detailed transaction history is available again, it is used automatically and takes priority, so the same spending is never counted twice.

Fixes #2257 

## Type of change


Georginio-prod:
image
image

Georginio-prod:
Capture d&#39;écran 2026-07-11 091039
Capture d&#39;écran 2026-07-11 091052

PR #2317, opened 2026-07-16
refactor: ♻️ use views instead of components for accounting routes

### LINKED ISSUE #2312: [Refactor] Improve page section for accounting
## Description

**Describe the bug**
Use views instead of component for routing page
Image

Georginio-prod:
# Description

Fixes #2312 — the accounting pages now follow the same routing structure as the rest of the app.

The accounting section was built differently from every other page: its routes pointed straight at section components and were rendered through a nested `RouterView` inside a wrapper view. This PR removes that wrapper and gives each accounting page (Summary, Income Statement, Balance Sheet, Trial Balance, General Ledger) its own view file under `views/team/[id]/Accounting/`, the way every other team page works. A small shared shell (`AccountingPage.vue`) keeps the header, the error banner and the single data load common to all five pages, so nothing changes for the user — URLs, sidebar links and breadcrumbs stay exactly the same.

Passes lint, format, type-check, the unit test suite and a production build in `app/`.

## Type of change


Georginio-prod:
image

PR #2319, opened 2026-07-19
feat: ✨ expandable accounting line items

### LINKED ISSUE #2249: Expandable Accounting Line Items
## Description
Currently it is very difficult to trust all of the lines in the accounting section.
Of course I can see the numbers and guess roughly at what each means but it is very difficult to know if the line is accurate without knowing what numbers went into it.

Image

The goal of this task is to add a small magnifying glass button beside each line item on all of the financial statements except for the general Ledger. When a user clicks his small magnifying glass, a pop up will open that shows them the list of every single General Ledger item that went into calculating this line item.

Image

The pop up window should look like the General Ledger, but only include those data points that went into calculating the final value. In this way a user can inspect any line item to understand how the final number was calculated.

Image

All line items on the Balance Sheet, Income Statement and Trial Balance should have this magnifying glass and this functionality.

At present I am unable to trust the numbers generated by the accounting functionality because I am unable to investigate and understand where these numbers came from.

I would like a small magnifying glass button beside each one of the line items. When a user clicks this button, a pop up window opens showing them all the general Ledger entries that went into calculating this line item.

An alternative to the pop up could be a drop down accordion. However, I think that the pop up is better because it can be paginated.

## Acceptance Criteria
- As a user of the CNC portal,
- I want to understand each line item of the financial statements,
- When I click on the magnifying glass beside a line item,
- I am able to see all of the General Ledger entries that were used to calculate this line item.

Georginio-prod:
# Description

## Initial Issue Description

Every figure on the accounting statements is a total, but there was no way to see *which* ledger entries add up to it. A number couldn't be checked without exporting the whole book and searching by hand.

Closes #2249

Each statement line that corresponds to a real account is now clickable. Clicking it (or its "Details" button) opens a popup listing exactly the ledger entries that make up that line's balance — paged, and reconciling the figure shown on screen. From the popup you can choose which columns to display and download the same view as PDF or Excel.

This works on all four statements:

- **Trial Balance** — every account row.
- **Income Statement** — each revenue and expense line.
- **Balance Sheet** — the individual assets and liabilities, each equity account, each cash pocket, and **Retained earnings** (which opens the combined ledger of every income and expense account it nets).

Lines that don't map to a single account — the "Cash (all pockets)" total and the per-currency sub-lines — stay non-clickable, so a click never shows a misleading figure.

Also included:

- A column-visibility picker inside every drill-down popup, remembered per statement.
- Consistent, quieter hover styling across the statements (rounded highlight, no divider bands).

## Type of change

**Please check if the PR fulfills these requirements**

- [x] Tests for the changes have been added

## Contribution checklist

Rules in [CONTRIBUTION.md](./../CONTRIBUTION.md) applied.


Georginio-prod:
image
image
image

PR #2330, opened 2026-07-21
feat: ✨ read accounting on-chain history directly from the blockchain

### LINKED ISSUE #2331: feat: ✨ read accounting on-chain history directly from the blockchain
## Description

The accounting page relied on a separate indexing service to read a team's on-chain history — deposits, withdrawals, transfers, wage settlements, share issuance, fees. Whenever that service was down, slow, or lagging behind the chain, the books came up incomplete or empty and the team had no way to see their accounting.

Read the on-chain history straight from the blockchain instead, so the accounting no longer depends on the indexing service being available. Every money source (Bank, payroll, expenses, investments, deposit router, and fees) is rebuilt the same way, and everything downstream — the ledger and the three financial statements — stays exactly the same.

Keeping the indexing service and adding fallbacks when it fails. Rejected: it keeps the hard dependency and the same failure mode instead of removing it.

**Additional context**
While reconstructing the history, a related display problem was also fixed: when several weekly pay claims were withdrawn on the same day, the hours shown on each settlement were wrong (they all showed the most recent claim's hours). Each settlement is now matched to the claim it actually paid, so the hours are correct.

## Acceptance Criteria

- [ ] The accounting page loads the full on-chain history without the indexing service.
- [ ] All money sources are covered: Bank, payroll, expenses, investments, deposit router, and fees.
- [ ] The ledger and the three financial statements show the same results as before.
- [ ] When several pay claims are withdrawn on the same day, each settlement shows the correct hours.
- [ ] The existing accounting tests pass.

Georginio-prod:
# Description

## Initial Issue Description

Fixes #2331 — the accounting page depended on a separate indexing service to read a team's on-chain history. When that service was down or lagging, the books came up incomplete or empty.

## PR Summary 

The accounting data layer now rebuilds each team's on-chain history **directly from the blockchain** instead of querying the indexing service. Per contract it reads the event logs from the deploy block, decodes them against the contract's multi-version ABI, dates each event from its block, and returns the **exact same shape** the indexer queries did — so every mapper and the statement engine are untouched. All five sources are migrated (Bank, CashRemuneration, Expense, InvestorV1, SafeDepositRouter), plus the global FeeCollector fees. The data-source swap lives in `useCNCAccounting`; the reusable base is `useContractEventsViaLogs`, which now also exposes `refetch` so a manual refresh re-scans the chain.

## Type of change

## Testing

- New unit test reproducing the same-day multi-claim settlement pairing.
- `app/` accounting suites pass: 263 accounting-util tests + 16 accounting-composable tests green.

### DIFF
--- a/app/src/utils/accounting/enrichment.ts
+++ b/app/src/utils/accounting/enrichment.ts
@@ -18,6 +18,7 @@ import { getAddress, isAddress, type Address } from 'viem'
 import type { TokenId } from '@/constant'
 import type { Wage, WeeklyClaim } from '@/types/cash-remuneration'
 import type { ExpenseResponse } from '@/types/expense-account'
+import { buildClaimRatesWithOvertime } from '@/utils/wageUtil'
 import type { LedgerEntry } from './ledgerEntry'
 
 export interface EnrichmentSources {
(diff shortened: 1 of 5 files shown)


app/src/utils/accounting/enrichment.ts
Georginio-prod, line 1: this files do update for writing of general ledger in accounting section about time how you submit

PR #2360, opened 2026-07-27
feat: ✨ show a running balance in the statement details

### LINKED ISSUE #2359: feat: ✨ show a running balance in the statement details
## Description

When you open the details behind a line of the trial balance, the income statement or the balance sheet, the popup lists the entries but gives a single figure at the end. There is no way to follow how the account got there: you cannot see what it was worth before the period, nor where it stands after each entry. Checking it against a bank statement or another bookkeeping tool means adding up the debit and credit columns by hand.

Show that detail the way a general ledger is normally read:

- a **Balance** column at the end, right after Credit, giving what the account stands at once each entry is booked
- an **Opening balance** line at the top, showing what the account brought in from before the period, with the movements behind it
- entries listed oldest first, so the balance builds up as you read down
- the foot of the table showing what the account is left with

A summary panel above the table with the totals (opening, debits, credits, closing). It gives the same figures but not the line-by-line progression, which is what makes the detail checkable.

**Additional context**
The balance only makes sense for a single account. The "Retained earnings" line of the balance sheet groups several accounts of different kinds, which share no common balance, so it keeps the current display.

## Acceptance Criteria

- [ ] The details popup of every statement (trial balance, income statement, balance sheet) shows a Balance column placed after Credit
- [ ] The balance is only filled on the lines of the account being opened; the other side of an entry stays empty
- [ ] The balance follows the nature of the account: it grows with debits for assets and expenses, with credits for liabilities, equity and income
- [ ] An "Opening balance" line heads the table with what the period carries in (zero when the period has no start date)
- [ ] Entries are listed oldest first and the last line shows the amount left at the end of the period
- [ ] Paging through the entries keeps the balance running from one page to the next
- [ ] A line that groups several accounts shows no Balance column
- [ ] Unit tests cover the calculation, the opening line and the display

Georginio-prod:
# Description

Fixes #2359 — the details behind a statement line now read like a general ledger, with the balance of the account followed line by line.

## Summary

Opening the details of a line in the trial balance, the income statement or the balance sheet used to give a flat list of the entries and a single figure at the end, so there was no way to see where the account stood at any point. This PR adds a **Balance** column at the end of that table, right after Credit, filled on the lines of the account being opened: it starts from an **Opening balance** line showing what the account brings in from before the period, moves with each entry according to the nature of the account (debits raise an asset or an expense, credits raise a liability, equity or income), and ends on the amount the account is left with, repeated at the foot of the table. Entries are now listed oldest first so the balance builds up as you read down, and paging keeps it running from one page to the next. The figures are taken from the debit and credit amounts actually shown, so the column can be checked by eye. A line that groups several accounts of different kinds — "Retained earnings" on the balance sheet — has no single balance and keeps the plain list. Along the way the ledger presenter had outgrown the size limit, so the shared column list moved into its own file with no call site changes. The general ledger tab is untouched.

## Type of change

## Tests

`build`, `lint`, `format`, `type-check` and the unit tests pass in `app/`. 15 new tests cover the calculation against the reference book (it ends on the account balance, follows the nature of the account, carries an opening amount forward, stays consistent across pages, leaves the facing side of an entry empty), the opening line, and the column appearing only for a single account.


Georginio-prod (commented):
lgtm

PR #2389, opened 2026-07-29
test: ✅ stop the unit suite from writing PDF exports into app/

### LINKED ISSUE #2388: test: ✅ stop the unit suite from writing PDF exports into app/
## Description

**Describe the bug**

Running the `app/` unit suite writes three real PDF files at the root of `app/`, untracked, on every local run:

```
app/Balance Sheet - As of <today>.pdf
app/Income Statement - All time.pdf
app/Trial Balance - As of <today>.pdf
```

They come from `AccountingView.spec.ts`: the three "exports and prints … from the export bar" tests click the real export buttons, which run the real `useAccountingExport` → `exportTablesPdf` → `doc.save(filename)`. jsPDF only knows how to trigger a browser download; the export settles after the spec file's jsdom window is gone, so it falls back to writing the file to the process cwd.

Nothing ignores those files either, so a `git add -A` in `app/` commits them by accident.

**To Reproduce**

1. `cd app`
2. `CI=1 npm run test:unit -- --run`
3. `ls *.pdf` — three PDFs, and `git status` shows them as untracked

**Expected behavior**

- The unit suite writes no files outside the test runner's own output.
- Export artifacts accidentally produced at the root of `app/` are git-ignored, so a future regression cannot be committed by mistake.

**Additional context**

- The two specs that already exercise the export path (`app/src/utils/__tests__/accountingPdf.spec.ts`, `app/src/composables/accounting/__tests__/useLedgerDrilldown.spec.ts`) mock their way around the writer; `AccountingView.spec.ts` is the one that does not.
- `app/src/utils/excelExport.ts` writes spreadsheets through the same download mechanism, so the ignore rule should cover `*.xlsx` at the root of `app/` too.
- Stubbing `jspdf` alone is not enough: the export resolves after the module registry is torn down, so one of the three writes slips past the mock. The composable boundary is the reliable place to stub.

## Acceptance Criteria

- [ ] `CI=1 npm run test:unit -- --run` in `app/` leaves no `.pdf` / `.xlsx` behind.
- [ ] The export-bar tests still assert that clicking Export/Print triggers the export with the expected filename.
- [ ] `app/*.pdf` and `app/*.xlsx` at the root of `app/` are git-ignored, without ignoring legitimate documents committed elsewhere in the repo.

hermannleboss:
# Description

## Intial Issue Description

Closes #2388

Running the `app/` unit suite dropped three real PDF files at the root of `app/`, untracked, on every local run:

```
app/Balance Sheet - As of <today>.pdf
app/Income Statement - All time.pdf
app/Trial Balance - As of <today>.pdf
```

**1. The write, at the source.** `AccountingView.spec.ts` was the only spec exercising the real export path: its three "exports and prints … from the export bar" tests click the real buttons, which run `useAccountingExport` → `exportTablesPdf` → `doc.save(filename)`. jsPDF only knows how to trigger a browser download, and the export settles *after* the spec file's jsdom window is torn down, so it took its Node fallback and wrote each PDF to the process cwd.

The spec now stubs `useAccountingExport`, so the click stops at the composable boundary. Worth noting for anyone hitting this again: stubbing `jspdf` itself is **not** enough — because the export resolves after the module registry is gone, one of the three writes reloads the real module and slips past the mock. The composable is the reliable seam.

Since the exports are stubbed, the three tests can assert something real instead of the tautological `expect(button.exists()).toBe(true)` they carried: each card must hand the export the filename its section is named after (`Trial Balance …xlsx` / `.pdf`, and so on). The builders behind the composable stay covered by `accountingPdf.spec.ts`, the filename rules by `exportNaming.spec.ts`.

**2. The safety net.** `/app/*.pdf` and `/app/*.xlsx` are now git-ignored — `app/src/utils/excelExport.ts` writes spreadsheets through the same download mechanism, so it can leak the same way. The rule is scoped to the root of `app/`, not a blanket `*.pdf`, so documents committed deliberately elsewhere in the repo still show up in `git status`.

Side benefit: the suite no longer renders three real PDFs, and local wall-clock dropped from ~110s to ~58s.

## Verification

Full `app/` quality gate, all green:

- `npx eslint .` — clean (one pre-existing warning in `registry.spec.ts`, untouched here)
- `npx prettier . --check` — clean
- `CI=1 npm run build-only` then `npm run type-check` — clean
- `CI=1 npm run test:unit -- --run` — 301 files, 2545 passed / 1 skipped, and `app/` is left with no `.pdf` / `.xlsx`

## Type of change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,8 @@ scripts/
 
 # Local working todolist shared between dev and AI agents (never committed)
 /todolist.md
+
+# Accounting export artifacts dropped at the root of app/ when an export runs
+# outside a browser (jsPDF / the Excel writer fall back to the process cwd)
+/app/*.pdf
+/app/*.xlsx
--- a/app/src/components/sections/AccountingView/__tests__/AccountingView.spec.ts
+++ b/app/src/components/sections/AccountingView/__tests__/AccountingView.spec.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest'
+import { describe, it, expect, vi, beforeEach } from 'vitest'
 import { flushPromises } from '@vue/test-utils'
 import { renderWithProviders } from '@/tests/mocks'
 import AccountingSummary from '../AccountingSummary.vue'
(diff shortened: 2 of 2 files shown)


hermannleboss (commented):
LGTM

PR #2391, opened 2026-07-30
docs: 📝 map contract read/write cache invalidation coupling

### LINKED ISSUE #2390: docs: 📝 map contract read/write cache invalidation coupling
## Context

There is no single place describing which on-chain reads a contract write invalidates. The result is
that every deposit / transfer component re-invents its own `queryClient.invalidateQueries` call, and
several of them target key shapes that match nothing.

The root cause is that `useContractWritesV3`'s built-in invalidation is narrower than its name
suggests: its predicate matches only keys shaped `['readContract', { address: <the contract written to> }]`.
That covers the view functions of the contract you just wrote to — and no balance of any kind:

- native balances live under `['balance', { address }]` (wrong `key[0]`);
- an ERC-20 balance of contract X is keyed on the **token** address, not on X.

Everything else (the `*-events-logs` transaction feeds, the hand-rolled Community Credit aggregates,
`['pastElections']`, …) is equally out of reach and needs an explicit `cfg.onSuccess`.

## Goal

Document the read/write invalidation coupling for the contract layer so that adding a write composable
is no longer guesswork:

- the three invalidation mechanisms and the `partialMatchKey` rule that governs them;
- an inventory of every on-chain read key namespace, and whether the built-in predicate reaches it;
- write → reads dirtied, per contract, with what covers each today;
- the reverse index (read → writes that dirty it);
- the recommended fix shape (per-domain invalidation helpers + shared key factories).

## Gaps found while mapping (follow-up work, not this issue)

- [ ] The 6 `*-events-logs` transaction feeds are never invalidated by anything (Bank, Expense,
      Community Credit, Investor, Cash Remuneration, SafeDepositRouter). A prefix predicate closes all six.
- [ ] `PayDividendsAction.vue` invalidates nothing after paying dividends.
- [ ] `TransferAction.vue` (ExpenseAccount `transfer`) invalidates only the backend expense list, no chain balance.
- [ ] The four `executeContractWrite` call sites bypass the built-in invalidation entirely
      (`useClaimMigration`, `useSweepMigration`, `useSetMigrationRoot`, `useShareholderMigration`).
- [ ] Cross-contract pairs with no invalidation: `ownerWithdrawAllToBank` → Bank balances,
      Vesting `release` → Investor balances, Elections `publishResults` → BoD membership,
      SafeDepositRouter `deposit` → Safe balance, Bank `fundFixedReturnRepayment` → both token balances.
- [ ] Dead keys that match no query: `['getBodActions']`, `['weekly-claims', teamId]`, `['team', { teamId }]`,
      and the `undefined`-laden key built by `useSyncWeeklyClaimsMutation`.
- [ ] `ProposalsList.vue` / `ContractOwnerCard.vue` read imperatively via `@wagmi/core`, so their data
      lives outside the query cache and no invalidation can reach it.
- [ ] A test that registers the real read keys against a real `QueryClient` and asserts each write's
      invalidation matches at least one of them. The current specs mock the query client and assert
      "was called with a key", which cannot catch a key that matches nothing.

hermannleboss:
# Description

## Intial Issue Description

Fixes #2390

There is no single place describing which on-chain reads a contract write invalidates, so every
deposit / transfer component re-invents its own `queryClient.invalidateQueries` call — and several
target key shapes that match nothing.

The root cause: `useContractWritesV3`'s built-in invalidation is narrower than its name suggests.
Its predicate matches only `['readContract', { address: <the contract written to> }]`, which covers
the view functions of that contract and **no balance of any kind** — native balances live under
`['balance', { address }]` (wrong `key[0]`), and an ERC-20 balance of contract X is keyed on the
**token** address, not on X. The `*-events-logs` transaction feeds, the Community Credit aggregates
and `['pastElections']` are equally out of reach.

Documentation only — no behaviour change.

`app/src/composables/contracts/INVALIDATION_MAP.md` (new):

1. **The three mechanisms** — the built-in V3 predicate, hand-written `onSuccess`, and "nothing" —
   plus the `partialMatchKey` rule that governs all three (`Object.keys(filter).every(...)` with an
   early `typeof a !== typeof b → false`).
2. **Read key namespaces** — the 10 on-chain key shapes in the app, and whether the built-in
   predicate can reach each one.
3. **Write → reads dirtied**, per contract (Bank, ERC-20, ExpenseAccount, CashRemuneration, Investor,
   FixedReturn, Vesting, Elections, BoD, SafeDepositRouter, Proposals, Safe), with the `file:line` of
   whatever covers each today.
4. **Reverse index** — read → the writes that dirty it, for debugging a stale value on screen.
5. **Ranked gaps** and the recommended fix shape (per-domain invalidation helpers, one key factory per
   namespace, feeds treated as part of a write's effect).

`README.md`: "What V3 gives you" step 4 previously read as if it invalidated every read of the
contract. It now states the predicate's real scope, names what falls outside it, and links the map.

## Notes for the reviewer

- The matching claims are not inferred from the docs — they were checked against the installed
  `@tanstack/query-core` `partialMatchKey`, and the vue-query `QueryClient` wrapper (which
  deep-unrefs filters, so refs inside a key are harmless and predicates survive the clone).
- The gaps found while mapping are tracked as a checklist in #2390 rather than fixed here, so the
  reference doc lands on its own and the fixes can be reviewed per domain.
- Worth a second opinion on the doc's home: it sits next to `composables/contracts/README.md` (which
  owns the V3 contract) rather than in `docs/contracts/`, on the assumption that whoever adds a write
  composable is already in that directory.

## Type of change

- [ ] Tests for the changes have been added (for bug fixes/features) — docs only; a real-`QueryClient`
      matching test is tracked in #2390

### DIFF
--- a/app/src/composables/contracts/INVALIDATION_MAP.md
+++ b/app/src/composables/contracts/INVALIDATION_MAP.md
@@ -1,0 +1,239 @@
+# On-chain invalidation map
+
+Which **contract reads** each **contract write** dirties, and by which mechanism (if any) they
+get invalidated today. Companion to [`README.md`](./README.md) (§"What V3 gives you").
+
+Scope: the chain layer only — `useReadContract` / `useBalance` / custom `useQuery` reads under
+`app/src/composables/**`. Backend REST invalidation is a separate surface (`app/src/queries/**`,
+see [`QUERY_FACTORY.md`](../../queries/QUERY_FACTORY.md)).
+
+**Read this before adding a write composable.** The built-in invalidation covers less than its
+name suggests, and the gap is not obvious from the call site.
+
+---
+
+## 1. The three mechanisms
+
+| #     | Mechanism                                                        | Where                                                    | What it reaches                                                                                                                |
+| ----- | ---------------------------------------------------------------- | -------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| **A** | Built-in predicate in `useContractWritesV3`                      | [`useContractWritesV3.ts:252`](./useContractWritesV3.ts) | queries whose key is `['readContract', { address: <written contract>, … }]` — scoped to `chainId` only when the write pins one |
+| **B** | `cfg.onSuccess` on the write, or an `onSuccess` at the call site | feature composables / components                         | whatever the author wrote by hand                                                                                              |
+| **C** | Nothing — the read refreshes on its own                          | —                                                        | `staleTime` expiry + refocus / remount, or an explicit `refetch()`                                                             |
+
+### The matching rule that decides everything
+
+`invalidateQueries({ queryKey })` uses `partialMatchKey` — `Object.keys(filter).every(...)` with an
+early `typeof a !== typeof b → false`. Two consequences that shape this whole document:
+
+1. **Mechanism A only fires on `key[0] === 'readContract'`.** Any other namespace is invisible to it.
+2. **An ERC-20 balance of contract X is keyed on the _token_ address, not on X.**
+   `balanceOf(bank)` → `['readContract', { address: USDC, args: [bank], … }]`. A write to the Bank
+   therefore **never** invalidates the Bank's token balances through mechanism A.
+
+Corollary: mechanism A covers _view functions of the contract you just wrote to_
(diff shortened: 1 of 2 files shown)


hermannleboss (commented):
Lgtm

PR #2393, opened 2026-07-30
build: 📦 generate typed as-const ABIs with @wagmi/cli

### LINKED ISSUE #2392: build: 📦 generate typed as-const ABIs with @wagmi/cli instead of hand-written `as Abi` wrappers
## Problem

The frontend's contract ABIs are hand-maintained wrappers of the form:

```ts
import Bank from './json/Bank.json'
export const BANK_ABI = Bank as Abi
```

A JSON import is widened by TypeScript (`string`, not `'function'`), so `as Abi` is the only way to
make it compile — and abitype can then derive nothing from it. Consequences today:

- **Write `args` are `readonly unknown[]`.** Wrong argument order, a `string` where the ABI says
  `uint256`, or a missing argument all compile fine and fail at simulation time, i.e. when the user
  clicks.
- **Read `data` is `unknown`,** which is why the codebase carries casts like
  `(isBodMember.value as boolean)` and `(await readContract(…)) as Address[]`.
- **A function name that does not exist on the contract compiles.** `useFixedReturnMarkAsRefundable`
  and `useFixedReturnClaimRefund` call `markAsRefundable` / `claimRefund`, neither of which exists on
  `FixedReturn.sol` (it exposes `refundLenders`).

`hardhat-abi-exporter` used to feed the `json/` output but was removed during the Hardhat 3 migration
(#2342) — it is a Hardhat 2 plugin (`peerDependencies: { hardhat: '^2.0.0' }`) and it only ever
emitted JSON, so it would not fix the typing anyway. `contract/scripts/copy-abi.ts` emits the right
`as const satisfies Abi` shape but carries a DRIFT WARNING telling readers not to run it: it covers
12 of 18 contracts under kebab names that don't all match the hand-written wrappers.

## Goal

Generate the ABIs `as const` with `@wagmi/cli`'s `hardhat` plugin so viem can resolve write `args`
tuples and read return types, and retire the hand-maintained wrappers.

- `contract/wagmi.config.ts` + `npm run generate-abi`, output committed to
  `app/src/artifacts/abi/generated.ts` (`contract/artifacts` is gitignored, so `app/` needs a
  checked-in copy to type-check in CI).
- Migrate every `@/artifacts/abi/<contract>` import and delete the superseded wrappers.
- Fix whatever the new typing reveals.

## Follow-ups (not this issue)

- [ ] Make `useContractWritesV3` generic over `abi` / `functionName` so `variables.args` resolves to
      the ABI tuple at every call site (needs dropping `MaybeRef` on those two fields — no call site
      passes a ref for either; the one dynamic ABI, `MainContractActions.vue`, gets an explicit
      escape hatch).
- [ ] Then narrow the per-domain write factories to `ContractFunctionName<abi, 'nonpayable' | 'payable'>`
      and drop `ExtractAbiFunctionNames`.
- [ ] Remove the ~44 now-unnecessary `as` casts on read data.
- [ ] Clean up `app/src/artifacts/abi/json/` (52 files, now only `FixedReturn.json` is still read, by
      `useFixedReturnEventsViaLogs` for its multi-version event union).

hermannleboss:
# Description

## Intial Issue Description

Fixes #2392

The frontend ABI wrappers were hand-maintained as `import X from './X.json'; export const X_ABI = X as Abi`.
A JSON import is widened by TypeScript (`string`, not `'function'`), so `as Abi` is the only way to make
it compile — and abitype can then derive nothing from it. Every write's `args` was `readonly unknown[]`
and every read's `data` was `unknown`.

The new typing found a real bug: **`useFixedReturnMarkAsRefundable` and `useFixedReturnClaimRefund`
called `markAsRefundable` / `claimRefund`, neither of which exists on `FixedReturn.sol`** — it exposes
`refundLenders(uint256)`. Any call would have failed at simulation. The comment above them claimed
"Issue Note's refund UI (OfferingIssuerActions.vue, OfferingDetail.vue) still calls them"; neither
component exists in the repo, and the only remaining references were their own specs and mock entries.
Removed, in `1b6b4ea4a`.

Same commit splits `FixedReturnFunctionNames`: one type covering every ABI function was fed to both
`useReadContract` and `useContractWritesV3`, so the read path accepted `transferOwnership` /
`initialize` / `addTokenSupport`. Now `ContractFunctionName<abi, 'view' | 'pure'>` for reads and
`'nonpayable' | 'payable'` for writes — which is also what makes `getLendingOffer` resolve to a real
return type instead of the union of every read, removing a bogus `LendingOfferStruct` cast in
`CreditLendModal.vue`.

Three commits, each independently green:

**`7c000458a` build** — `@wagmi/cli` as a `contract/` devDependency, `contract/wagmi.config.ts`,
`npm run generate-abi`, and the generated `app/src/artifacts/abi/generated.ts` (13 ABIs). Retires
`contract/scripts/copy-abi.ts`, which carried a DRIFT WARNING telling readers not to run it (12 of 18
contracts, kebab names that didn't all match the wrappers, would clobber them). `AGENTS.md` and the
Solidity audit checklist updated to the new command.

**`71a05c0b1` refactor** — ~85 imports migrated across 63 files; the 16 superseded wrappers deleted.
Collapses duplication they had accumulated: `bod.ts` / `board-of-directors.ts` exported the same
BoardOfDirectors ABI under two names, as did `investors.ts` / `investorV2.ts` (both read
`Investor.json`, so the v1/v2 split was already fictional at the ABI level) and `expense-account.ts` /
`expense-account-eip712.ts`; `board-of-directors.ts` and `expense-account.ts` had no importers at all.
ERC-20 now uses viem's own `erc20Abi`. `ownable-pausable.ts` stays hand-written — synthetic surface,
no `.sol` to compile — but becomes an `as const` literal.

**`1b6b4ea4a` fix** — the FixedReturn findings above.

## Notes for the reviewer

- **Hardhat 3 compatibility was verified, not assumed.** The plugin globs `artifacts/contracts/**` and
  doesn't touch Hardhat internals; the only adaptation is `commands.build` (HH3 renamed `compile` to
  `build`). `hardhat-abi-exporter` is not an option here — it is `peer: hardhat ^2.0.0`, was removed in
  #2342, and only ever emitted JSON.
- **The typing gain is measured.** A probe asserting the four mistakes with `@ts-expect-error` compiles
  clean against the generated ABI; the same probe against `bankAbi as Abi` reports all three directives
  as *unused*, i.e. today's wrappers catch none of them.
- **`include` is explicit on purpose.** Without it the plugin also emits every interface, mock and test
  util — 40 ABIs / 8.7k lines instead of 13 / 6.1k. Adding a contract to the frontend means adding it
  to that list.
- **The generated file is committed** because `contract/artifacts` is gitignored. Single-file output is
  the CLI's default; happy to split per contract if you'd rather keep diffs narrow.
- **No TS bump needed.** `@wagmi/cli` peers `typescript >=5.7.3` and lives in `contract/` (TS ^6.0.3);
  the generated file is plain `as const` arrays and compiles under `app/`'s TS ^5.6.2.
- `app/src/artifacts/abi/json/` (52 files) is now near-orphaned — only `FixedReturn.json` is still read,
  deliberately, by `useFixedReturnEventsViaLogs` for its multi-version event union. Cleanup left to its
  own PR (tracked in #2392).

## Verification

- `npm run type-check` clean. 10 pre-existing `TS7006` remain (UTable slot params in Community Credit);
  confirmed identical on a clean `develop` by stashing and re-running.
- `npm run test:unit` — 304 files, 2578 tests, all pass.
- `npm run lint`, `npm run format-check`, prettier on `contract/` + docs, and
  `scripts/audit-doc-drift.sh` all green.

## Type of change

- [x] Tests for the changes have been added (for bug fixes/features) — existing suite covers the
      migrated paths; the two removed FixedReturn wrappers had their specs removed with them

## Large PR justification

634 production lines, over the 400 cap. Almost all of it is one mechanical substitution repeated 63
times:

```diff
-import { BANK_ABI } from '@/artifacts/abi/bank'
+import { bankAbi } from '@/artifacts/abi/generated'
```

counted twice per file (a deletion and an addition), plus the ~80 lines of the 16 deleted wrapper files.

**Why splitting by domain doesn't help.** The wrappers are deleted in the same commit their consumers
move — that is what makes the rename atomic. Splitting per contract would mean 13 PRs of a single
rename, or leaving dead wrappers alive across several PRs.

**The alternative I considered and rejected:** turn each wrapper into a re-export of the generated
const (`export const BANK_ABI = bankAbi`). That lands the typing in ~40 lines and defers all the
churn — but it adds a transitional indirection layer that then needs its own PR to remove, so it is
more total churn for the same end state. Happy to redo it that way if you'd rather review it in two
passes; say so and I'll restructure.

**What actually needs your attention** (the rest is the substitution above):

- `contract/wagmi.config.ts` — 48 lines. In particular the `include` list, which has to be extended
  whenever the frontend starts consuming a new contract.
- `1b6b4ea4a` — the FixedReturn fix, +29/−59. This is the only behaviour change in the PR.
- `app/src/artifacts/abi/ownable-pausable.ts` — rewritten by hand as an `as const` literal; worth
  checking the entries match what `contractManagementUtil.ts` calls.

Reviewing commit by commit (`7c000458a` → `71a05c0b1` → `1b6b4ea4a`) separates the tooling, the
mechanical rename, and the fix.


hermannleboss (commented):
LGTM

PR #2400, opened 2026-08-01
refactor: ♻️ type contract write args against the ABI

### LINKED ISSUE #2399: refactor: ♻️ type contract write `args` against the ABI in useContractWritesV3
Follow-up to #2392, which generated the ABIs `as const` but left the write path untyped.

## Problem

`useContractWritesV3` widens `abi` to `Abi` and declares `args?: readonly unknown[]`, so nothing
about a write call is checked against the function signature:

```ts
const transfer = useTransferToken()   // transferToken(address token, address to, uint256 amount)

await transfer.mutateAsync({ args: [to, token, amount] })  // wrong order   → compiles
await transfer.mutateAsync({ args: [token, to, '1000'] })  // string/uint256 → compiles
await transfer.mutateAsync({ args: [token, to] })          // missing arg   → compiles
```

All three fail at simulation time, i.e. when the user clicks. Reads are already fully typed —
`useReadContract` is generic and only needed the `as const` ABIs that #2392 delivered.

Worth noting the per-domain factories' function-name check was **vacuous** before #2392:
`ExtractAbiFunctionNames<Abi>` resolves to `string`, so `useBankContractWrite('anythingAtAll')`
type-checked.

## Goal

Make `useContractWritesV3` generic over `abi` / `functionName` so `args` resolves to the tuple the
ABI declares, and narrow the per-domain factories to match.

This requires `abi` and `functionName` to become plain values rather than `MaybeRef`: a reactive
`functionName` makes the arg derivation the union of every function's tuple, which constrains
nothing. `contractAddress` and `chainId` stay reactive — they don't feed the arg types. No call site
passes a ref for either narrowed field.

## Follow-ups (not this issue)

- [ ] The three read helpers in `composables/fixedReturn/reads.ts` take `functionName` as a
      non-generic parameter, so their return type degrades to the union of every view function's
      return (`useFixedReturnOwner()` yields `string | bigint | readonly string[] | …` instead of
      `Address`). One line each: `<F extends FixedReturnReadNames>(functionName: F)`.
- [ ] Remove the read-side `as` casts that the typed ABIs made unnecessary
      (`(isBodMember.value as boolean)`, `(await readContract(…)) as Address[]`, …).
- [ ] Clean up `app/src/artifacts/abi/json/` — only `FixedReturn.json` is still read, by
      `useFixedReturnEventsViaLogs` for its multi-version event union.

hermannleboss:
# Description

## Intial Issue Description

Fixes #2399 — follow-up to #2392, which generated the ABIs `as const` but left the write path
untyped.

`useContractWritesV3` widened `abi` to `Abi` and declared `args?: readonly unknown[]`, so a wrong
argument order, a `string` where the ABI says `uint256`, or a missing argument all compiled and
failed at simulation time — when the user clicks. Reads were already fully typed; they only needed
the `as const` ABIs.

The new typing surfaced two real defects, both committed **before** the refactor so every commit is
independently green:

**`b17e423a7` — `BoardOfDirectors.addAction` could be called with undefined arguments.**
`executeAddAction` took `Partial<Action>`, so `targetAddress` / `description` / `data` were
optional, and the args tuple was cast to `readonly unknown[]`. I checked what viem actually does
here: `encodeFunctionData` **encodes `undefined` rather than throwing**, so an incomplete call would
have submitted a governance action with an empty target and payload, silently, instead of failing.
All four call sites already pass the three fields.

**`79261092a` — `Elections.createElection` received `number` for `uint256` params.** Also checked:
viem accepts numbers for uint types, so this was not broken in production — but it silently loses
precision above 2^53, and it is what forced the `as readonly unknown[]` cast on the args array.

**`2699d1445`** makes `useContractWritesV3` generic over `abi` and `functionName`, so viem resolves
`args: ContractFunctionArgs<abi, 'nonpayable' | 'payable', fn>`.

`abi` and `functionName` become plain values rather than `MaybeRef`. That is the trade that buys the
typing: a reactive `functionName` makes the arg derivation the union of every function's tuple,
which constrains nothing. `contractAddress` and `chainId` stay reactive — neither feeds the arg
types — and no call site passed a ref for either narrowed field.

**The one apparently dynamic ABI was not dynamic.** `MainContractActions.vue` did
`computed(() => props.row.abi as Abi)`, but `getTeamContracts` sets `abi: ownablePausableAbi` on
every row — so it was that single ABI laundered through an untyped `TableRow` (`Record<string, any>`)
and re-widened. It now imports the ABI directly, and the escape hatch I expected to need doesn't
exist.

Per-domain factories become generic pass-throughs using the exported `WriteFunctionName<abi>`, which
also makes their function-name check real for the first time: `ExtractAbiFunctionNames<Abi>` resolves
to `string`, so before #2392 `useBankContractWrite('anythingAtAll')` type-checked.
`elections/writes.ts` was additionally constraining on `ElectionsFunctionName`, which covers reads.

`WriteFunctionName` and `WriteFunctionArgs` are exported for helpers that forward args;
`TransferAction.vue`'s `submitExpenseAccountTransfer` is the first user.

## Notes for the reviewer

- **One deliberate cast remains**, in `useContractWritesV3`'s `mutationFn`: `executeContractWrite` is
  the framework-agnostic escape hatch and stays loosely typed, so the tuple is widened at that
  boundary after being validated against the ABI above it.
- **`develop` is currently red** — 5 unit tests fail in
  `CommunityCreditView/__tests__/CreditCallTermsStep.spec.ts` and
  `CommunityCredit/__tests__/NewView.spec.ts`. I verified this on a clean `develop` by stashing;
  they are not from this PR and I have not touched them. They go unnoticed because `app-unit-test`
  is not a required check on develop — only `codecov/patch` and `codecov/project` are.
- **Still open after this PR:** the three read helpers in `composables/fixedReturn/reads.ts` take a
  non-generic `functionName`, so their return type degrades to the union of every view function's
  return. One line each; tracked in #2399.

## Verification

- `npm run type-check` — 0 errors.
- `npm run test:unit` — 2608 pass, 5 fail (the pre-existing `develop` failures above; no new ones).
- `npm run lint` and `npm run format-check` — clean.

## Type of change

- [x] Tests for the changes have been added (for bug fixes/features) — the guarantees here are
      compile-time; the existing suite covers the touched call sites
- [x] Docs have been added / updated (for bug fixes / features) — `composables/contracts/README.md`
      said `args` was "**not** structurally validated"; that paragraph is now the opposite

### DIFF
--- a/app/src/components/sections/AdministrationView/CurrentBoDElectionSection.vue
+++ b/app/src/components/sections/AdministrationView/CurrentBoDElectionSection.vue
@@ -115,15 +115,15 @@ const createElection = async (electionData: OldProposal) => {
       electionData.title,
       electionData.description,
       dateToUnixTimestamp(electionData.startDate as Date) < dateNow
-        ? dateNow + 60 // Start in 1 minute if start date is in the past
-        : dateToUnixTimestamp(electionData.startDate as Date),
+        ? BigInt(dateNow + 60) // Start in 1 minute if start date is in the past
+        : BigInt(dateToUnixTimestamp(electionData.startDate as Date)),
       dateToUnixTimestamp(electionData.startDate as Date) < dateNow
-        ? dateNow + 60 + 60 // End 1 minute after adjusted start time if start date is in the past
-        : dateToUnixTimestamp(electionData.endDate as Date),
-      electionData.winnerCount,
+        ? BigInt(dateNow + 60 + 60) // End 1 minute after adjusted start time if start date is in the past
+        : BigInt(dateToUnixTimestamp(electionData.endDate as Date)),
+      BigInt(electionData.winnerCount),
       electionData.candidates?.map((c) => c.candidateAddress) || [],
       teamStore.currentTeam?.members.map((m) => m.address) || []
-    ] as readonly unknown[]
+    ] as const
 
     await writeCreateElection({ args })
 
--- a/app/src/components/sections/ContractManagementView/MainContractActions.vue
+++ b/app/src/components/sections/ContractManagementView/MainContractActions.vue
@@ -91,7 +91,8 @@
 </template>
 <script setup lang="ts">
 import { Icon as IconifyIcon } from '@iconify/vue'
-import { encodeFunctionData, type Abi, type Address } from 'viem'
+import { encodeFunctionData, type Address } from 'viem'
+import { ownablePausableAbi } from '@/artifacts/abi/ownable-pausable'
 import type { TableRow } from '@/types/table'
 import { ref, computed, watch } from 'vue'
 import { useTeamStore, useUserDataStore } from '@/stores'
@@ -167,7 +168,6 @@ const formatedActions = computed(() => {
 })
 
 const rowAddress = computed(() => props.row.address as Address)
-const rowAbi = computed(() => props.row.abi as Abi)
 
 const {
   mutate: executeTransferOwnership,
@@ -175,7 +175,7 @@ const {
   error: errorTransferOwnership
 } = useContractWritesV3({
   contractAddress: rowAddress,
-  abi: rowAbi,
+  abi: ownablePausableAbi,
   functionName: 'transferOwnership'
 })
 
@@ -185,7 +185,7 @@ const {
   error: errorPauseContract
 } = useContractWritesV3({
   contractAddress: rowAddress,
-  abi: rowAbi,
+  abi: ownablePausableAbi,
   functionName: 'pause'
 })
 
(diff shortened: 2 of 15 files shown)


hermannleboss:
`app-unit-test` is red here for the five pre-existing Community Credit failures described in the PR body, not for anything in this branch (CI reports `2 failed | 306 passed`, the same as a clean `develop`).

I traced them: they are not flaky or stale specs but a live bug — `createDefaultCreditCallForm` hard-codes `deadline: '2026-07-31'`, so since 2026-08-01 the wizard pre-fills a deadline in the past and refuses to publish. Fixed in #2402, which branches from `develop` and is independent of this PR. This one goes green once that lands.

PR #2406, opened 2026-08-01
fix: 🐛 recognise the current Investor contract type, not just InvestorV1

### LINKED ISSUE #2405: fix: 🐛 lookup tables keyed on InvestorV1 skip the current Investor contract
## Description

**Describe the bug**

Since the v2 share token shipped, a team's Investor contract is registered as `Investor`, while teams on legacy Officer generations still carry `InvestorV1`. Several lookup tables were keyed on `InvestorV1` alone and never gained an `Investor` entry, so they silently degrade for every team on a current Officer:

1. **`INTERNAL_POCKET_CONTRACT_TYPES`** (`app/src/utils/accounting/internalAddresses.ts`) decides which of a team's contracts count as its own money pockets. With the current share token absent, accounting classifies it as an **external counterparty** and books its transfers as third-party movements.
2. **`CONTRACT_ABI_MAP`** (`app/src/utils/abiDecodeUtil.ts`) resolves the ABI used to decode calldata in the transaction detail modal. With no `Investor` key, share-token transactions cannot be decoded at all.
3. **The revert-message catalog** (`app/src/composables/contracts/errorCatalogs.ts`) has an `InvestorV1` contract key but no `Investor`, so v2 reverts fall through to the generic `Transaction failed`.

Separately, naming that no longer matches reality:

- The snapshot endpoint's body field is `investorV1Address`, but the migration source is the previous Officer's share token — any generation (V0 / V0.1 / V1), or an `Investor` on a V2→V2 redeploy. The sibling persistence endpoint and the `InvestorMigration` model already call it `previousInvestorAddress`.
- `app/src/artifacts/abi/V2/investors.ts` exports the frozen **InvestorV1** ABI under the name `INVESTOR_ABI` from the **V2** folder. Nothing imports it, and v2 ABIs come from `generated.ts`.
- Several comments describe the *current* share token as an `InvestorV1`. Vesting in particular is a v2-only contract, so it can never mint from an `InvestorV1`.

**To Reproduce**

1. Use a team whose Officer is a current generation (its share token is registered as `Investor`, not `InvestorV1`).
2. Open the accounting view — the share token address is treated as external, so movements to and from it are booked as third-party.
3. Open a share-token transaction in the detail modal — the calldata is not decoded.
4. Trigger a failing claim on the shareholder migration page — the message is `Transaction failed`, with no indication of which guard rejected it.

**Expected behavior**

Both share-token generations resolve everywhere a contract type is looked up, and migration reverts produce an actionable message. Naming reflects that the migration source is "the previous Investor", not specifically a v1.

**Additional context**

The revert catalog is also missing the migration reverts entirely — `InvalidProof`, `MigrationRootNotSet`, `MigrationAlreadyComplete`, `AlreadyMigrated`, `LengthMismatch`, `DividendsFrozenDuringMigration`. These surface on the shareholder claim page, which is exactly where a bare `Transaction failed` leaves a stuck holder with nothing to act on.

Mentions of `InvestorV1` that name the **legacy** contract are correct and stay: the `TeamContract` type union, the backend contract-type enum, the legacy ABI folders, the `getTeam()` lookups that accept both strings, and the ponder handlers indexing real legacy contracts.

## Acceptance criteria

- [ ] A team on a current Officer has its share token recognised as an internal accounting pocket.
- [ ] Share-token transactions decode in the detail modal for both generations.
- [ ] Migration reverts resolve to a specific, actionable message.
- [ ] No identifier or comment claims the current share token is an `InvestorV1`.
- [ ] Legacy `InvestorV1` references that name the real legacy contract are preserved.

hermannleboss:
# Description

## Intial Issue Description

Fixes #2405

Several lookup tables were keyed on the `InvestorV1` contract type alone and never gained an `Investor` entry when the v2 share token shipped, so they silently degrade for every team on a current Officer generation. Plus naming that no longer matches reality now that the migration source can be any legacy generation.

Three commits, each one logical change.

**`refactor: ♻️ rename the snapshot source param to previousInvestorAddress`** — `investorV1Address` claimed the migration source is always an InvestorV1. It is not: the previous Officer can be V0, V0.1 or V1, and a V2→V2 redeploy reads an `Investor`. Renamed across the snapshot route, controller, service and the frontend caller, which also aligns it with the persistence endpoint and the `InvestorMigration` model that already used that name. **The request-body field of `POST /investor-migration/snapshot` changes**; no stored data is affected.

**`fix: 🐛 recognise the current Investor contract type, not just InvestorV1`** — the substantive part:

- `INTERNAL_POCKET_CONTRACT_TYPES` — the current share token was classified as an *external* counterparty, so accounting booked its transfers as third-party movements.
- `CONTRACT_ABI_MAP` — no `Investor` key meant share-token transactions could not be decoded in the detail modal at all.
- The revert catalog key, so v2 reverts fell through to the generic `Transaction failed`.

The catalog key is *renamed* rather than duplicated: the error names it maps (`NotBank`, `NoShareholders`, …) are shared by both contracts and no production caller passed `'InvestorV1'`. The other two keep **both** keys, since legacy teams still carry `InvestorV1`. The catalog also gains the migration reverts (`InvalidProof`, `MigrationRootNotSet`, `MigrationAlreadyComplete`, `AlreadyMigrated`, `LengthMismatch`, `DividendsFrozenDuringMigration`) — they surface on the shareholder claim page, exactly where a bare `Transaction failed` leaves a stuck holder with nothing to act on. Covered by new tests, including a regression test asserting the current-generation Investor is an internal pocket.

**`docs: 📝 stop calling the current share token InvestorV1`** — comments describing the *current* token as an InvestorV1 (Vesting is v2-only, so it can never mint from one). Mentions describing the *previous* contract are left alone, and the two `getTeam()` lookups gain the reason they accept both strings: every legacy generation registers its share token as `InvestorV1`, current ones as `Investor`, and `getShareholders` is ABI-identical across all of them. Also deletes the unused `artifacts/abi/V2/investors.ts`, which exported the frozen **InvestorV1** ABI under the name `INVESTOR_ABI` from the **V2** folder — a trap for whoever reached for it first.

**`chore: 🔧 delete the unused typed ABI wrappers frozen under abi/V2`** — none of the 17 `.ts` wrappers in `app/src/artifacts/abi/V2/` is imported anywhere. Live ABIs come from the top-level `generated.ts` (58 importers) and frozen generations are read through their `json/` folders, which is the interface `composables/contracts/README.md` documents. They are the fossil of a superseded pattern: `freeze-version.ts` copies every sibling `.ts` of `artifacts/abi/`, and V2 was frozen while the app still had one wrapper per contract — the absence of a `generated.ts` in that folder dates the snapshot. Being dead is the smaller problem: they are pinned to the artifacts of their freeze day, so they drift from `generated.ts` on every recompile while keeping export names credible enough to be picked up by autocompletion. That already bit once — the `V2/investors.ts` removed earlier in this branch served the *InvestorV1* ABI under the name `INVESTOR_ABI`, from the V2 folder. `V2/json/` is untouched.

## Contribution

Deliberately **not** renamed: `InvestorV1` where it names the real legacy contract — the `TeamContract` type union, the backend contract-type enum, the legacy ABI folders and their beacon/Ignition keys, the `getTeam()` lookups, and the ponder handlers indexing real legacy contracts on-chain. Those are data, not naming debt.

Left for a follow-up that touches `contract/scripts/`: the same dead wrappers in `ponder/abis/V2` (12) and `dashboard/app/artifacts/abi/V2` (1), plus restricting the `freeze-version.ts` filter to `json/` so the next freeze stops recreating them.

## Type of change


hermannleboss (commented):
LGTM

PR #2412, opened 2026-08-02
refactor: ♻️ drive useContractBalance from TanStack Query and wagmi core

### LINKED ISSUE #2411: refactor: ♻️ drive useContractBalance from TanStack Query and wagmi core
## Context

`useContractBalance` builds one wagmi composable per supported token inside a `.map()` — a `useBalance` for native, a `useReadContract` for each ERC-20. That shape causes several problems:

- **Composables created in a loop.** The address is captured once via `unref()`, so the balance never follows a reactive address. The file carries a standing `// TODO: support reactive address changes`.
- **N queries per address**, each with its own loading and error state. `isLoading` is an `.some()` across them and `error` is whichever one failed first.
- **A failing read is silent.** The token whose read failed reports an amount of 0, so the displayed total is wrong rather than absent.
- **A balance that is loading is indistinguishable from a balance of 0** — every consumer receives a full array of zeroed entries until the data lands.

The returned data structure has its own issues:

- `TokenBalance.values` is a `Record<currencyCode, …>`, but the currency store only ever emits two price rows, identified by `id`: `local` and `usd`. When the local currency **is** USD — the default — both rows carry `code: 'USD'` and the second overwrites the first. `SafeBalanceSection` carries a `?? total['USD']?.formated ?? 0` fallback that exists purely to cover that missed lookup.
- Three of the seven fields on each value object (`id`, `code`, `symbol`) have no reader anywhere in the app, and `code` duplicates the map key.
- A total inherits `price` / `formatedPrice` from the *first token*, which means nothing for an aggregate.
- Amounts are only exposed as `number`, built from up to 18 decimals, yet transfer and deposit forms validate against them.

## Goal

Drive the composable from TanStack Query directly, on top of `@wagmi/core` functions, and return the query itself.

- One query per address covering every token, issued in a single tick so the batching transport collapses the reads into one JSON-RPC round-trip.
- Cache the raw `bigint` amounts; derive prices in a `computed`, so switching currency re-prices without a refetch.
- `data` is `undefined` until the first successful read, per TanStack semantics.
- Key balance amounts by currency role (`usd` / `local`) rather than by currency code.
- Expose the exact on-chain `bigint` alongside the decimal amount.
- Keep the existing invalidation contract working, and give the key a factory so call sites stop hand-writing wagmi's key shape.

## Scope

- `app/src/composables/useContractBalance.ts` and a new pure `app/src/lib/balances/` layer
- `app/src/types/contract-balance.ts`
- The ~16 view consumers, their specs, and the shared test mocks
- The shared invalidation sites (`useSafeSendTransaction`, `useCashOutAll`, `safeKeys`)

hermannleboss:
# Description

## Intial Issue Description

Fixes #2411

`useContractBalance` built one wagmi composable per token inside a `.map()` — a `useBalance` plus one `useReadContract` per ERC-20. The address was captured once through `unref()` (hence the standing `// TODO: support reactive address changes`), each token carried its own loading and error state, and a failing read reported an amount of 0 rather than an error, quietly producing a wrong total.

The returned shape had a defect of its own: `values` was keyed by currency **code**, but the currency store only ever emits two price rows identified by `id` — `local` and `usd`. When the local currency *is* USD (the default), both rows carry `code: 'USD'` and the second overwrote the first. `SafeBalanceSection`'s `?? total['USD']?.formated ?? 0` fallback existed purely to paper over that missed lookup.

The composable is now one TanStack query per address, reading through `@wagmi/core` directly, and it returns **the query itself**.

```ts
const { data, isLoading, error, refetch } = useContractBalance(addr)

data.value // ContractBalances | undefined
// {
//   balances: [{ token, raw: 12500000n, amount: 12.5,
//                price: { usd: {value,formatted}, local: {…} },
//                value: { usd: {…}, local: {…} } }, …],
//   total:    { usd: { value: 2012.5, formatted: '$2.01K' }, local: {…} }
// }
```

**One query, one round-trip.** All token reads are issued in the same tick, so the batching http transport collapses them into a single JSON-RPC POST. One loading state, one error state. A failing read now fails the query instead of reporting a silent 0 and a wrong total.

**`data` is priced but reactive.** The query caches raw `bigint`s; `data` is a `computed` over them. `useQuery` returns `toRefs(...)` — an object of refs — so the query can be spread and `data` replaced without losing reactivity. Switching currency re-prices from cache with no refetch; a test asserts `getBalance` is not called again.

**`data` is `undefined` until the first read lands**, per TanStack semantics. A balance that is loading is no longer indistinguishable from a balance of 0.

**Amounts are keyed by currency role, not code.** `{ usd, local }` cannot collide, so call sites lose the optional chaining and the fallbacks on the currency dimension:

```diff
- total[currency.code]?.formated ?? total['USD']?.formated ?? 0
+ data?.total.local.formatted ?? 0
```

Fields with no reader anywhere (`id`, `code`, `symbol`) are gone, `code` no longer duplicates the map key, and a total no longer carries a unit price inherited from the first token. `raw` (the exact on-chain `bigint`) is exposed so "max" buttons and amount validations can stop comparing floats rebuilt from 18 decimals — `OwnerTreasuryWithdrawAction` already uses it.

Fetching and currency derivation moved to `app/src/lib/balances/tokenBalances.ts` as pure functions, unit-tested without Vue.

Three latent bugs surfaced and were fixed along the way:

- **`chainId` passed as a ref into invalidation keys.** `DepositBankForm` and `DepositSafeForm` put the `Ref` itself in the key object, so `partialDeepEqual` compared a `Ref` against a `number` and those invalidations never matched anything.
- **The global module mock swallowed `contractBalanceKeys`.** `vi.mock('@/composables/useContractBalance', () => ({ useContractBalance }))` returned only the function, leaving the constant `undefined`; `contractBalanceKeys.detail(...)` threw a `TypeError` the component's `catch` absorbed, so the invalidation silently never fired and the test still passed. Now goes through `importOriginal`.
- **`GenericTokenHoldingsSection` sorted on fields the row never had.** The `price` and `balance` accessors pointed at properties absent from `TokenBalance`, so sorting was a no-op. Rows are built in a computed with sortable scalars.

Worth knowing: `tsconfig.app.json` excludes `__tests__` and `*.spec.ts`, so spec fixtures are **not** type-checked — several carried the old shape without any compile error. The new shared `makeTokenBalance` derives `raw` and `value` from the amount and price, so a fixture can no longer contradict itself.

## Contribution

Four atomic commits, each reviewable on its own:

| commit | scope |
|---|---|
| `7239efb7b` | composable → TanStack Query + `@wagmi/core`, pure layer in `lib/balances/` |
| `2720ff9e8` | data structure keyed by currency role |
| `f725f59ae` | 16 view consumers, specs, shared test mocks |
| `aaa71ab4f` | invalidations routed through `contractBalanceKeys` |

**Invalidation contract.** The query key deliberately keeps wagmi's `['balance', { address, chainId }]` shape, because a dozen call sites invalidate it by hand after a transfer, deposit or withdrawal — they keep working untouched. A `contractBalanceKeys` factory is exported so new call sites stop hand-rolling the literal, and `safeKeys.balance` now delegates to it. `safeKeys.tokenBalance` is removed: it built a per-ERC-20 `readContract` key that no longer backs any balance display, and both Safe mutations already invalidated `safeKeys.balance` alongside it.

**One judgement call worth a reviewer's eye:** `SafeBalanceSection.transfer.spec.ts` had an ETH fixture with `id: 'ethereum'`, which is not a valid `TokenId` — the old fixture was untyped. It is now `'native'`, and the `tokenId` assertion follows. If an `'ethereum'` token id genuinely circulates on the Safe path, that is a bug elsewhere; the typing says it does not.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Breaking in the sense that the composable's public shape changed; every consumer in the repo is migrated in this PR.

## Verification

- `npm run build-only` then `npm run type-check` — 0 errors
- `npm run test:unit` — 311 files, 2646 passed, 1 skipped
- `prettier --check src` — clean
- `eslint src` — 0 errors (1 pre-existing warning in an unrelated spec)

## Large PR justification

This changes the public shape of a composable used by 16 components, so the migration is atomic by construction: **every intermediate split leaves the repo unable to type-check.**

I did try the suggested decompositions:

- **Stack PRs** — the natural base would be commits `7239efb7b` + `2720ff9e8` (composable and data structure). Merged to `develop` on its own, that base breaks the build: all 16 consumers still destructure `{ balances, total }`, which no longer exists. A stack whose base cannot be merged independently is not a stack, it is this PR with extra review surface.
- **Extract prep work** — the new pure layer (`lib/balances/tokenBalances.ts`) and the shared `makeTokenBalance` fixture builder *could* land first, but they are 130 production lines with no caller until the composable switches over. Landing dead code first does not reduce what a reviewer has to hold in their head; it spreads it across two PRs.
- **Separate refactors from behavior changes** — there is no behavior change to separate out. The three bug fixes listed above are one-line consequences of the same shape change (a ref that had to be unwrapped, a mock factory that had to reexport, table accessors that had to point at real fields).
- **Split by layer** — everything here is `app/`.

On the numbers: the 843 test lines are almost entirely fixture rewrites forced by the same shape change, and they are a **net reduction** — the diff is 294 insertions against 473 deletions overall. Three specs had grown their own hand-rolled balance builders; those are replaced by one shared `makeTokenBalance`.

The four commits are individually reviewable and are best read in order — each has a message explaining its own rationale.

### DIFF
--- a/app/src/components/GenericTokenHoldingsSection.vue
+++ b/app/src/components/GenericTokenHoldingsSection.vue
@@ -3,20 +3,7 @@
   <UCard>
     <template #header>Token Holding</template>
     <UTable
-      :data="
-        balances.map((balance, index) => ({
-          ...balance,
-          rank: index + 1,
-          icon:
-            balance.token.symbol === 'USDC' || balance.token.symbol === 'USDCe'
-              ? USDCIcon
-              : balance.token.symbol === 'POL'
-                ? MaticIcon
-                : balance.token.symbol === 'ETH'
-                  ? EthereumIcon
-                  : null
-        }))
-      "
+      :data="rows"
       :loading="isLoading"
       :columns="[
         { accessorKey: 'rank', header: 'RANK' },
@@ -30,11 +17,11 @@
         {{ row.amount }} {{ row.token.symbol }}
       </template>
       <template #price-cell="{ row: { original: row } }">
-        {{ row.values[currency.code ?? 'USD']?.formatedPrice }} / {{ row.token.symbol }}
+        {{ row.priceLabel }} / {{ row.token.symbol }}
       </template>
 
       <template #balance-cell="{ row: { original: row } }">
-        {{ row.values[currency.code ?? 'USD']?.formated }}
+        {{ row.balanceLabel }}
       </template>
 
       <template #token-cell="{ row: { original: row } }">
@@ -57,22 +44,38 @@
 import EthereumIcon from '@/assets/Ethereum.png'
 import USDCIcon from '@/assets/usdc.png'
 import MaticIcon from '@/assets/matic-logo.png'
+import { computed } from 'vue'
 import { useContractBalance } from '@/composables'
-import { useStorage } from '@vueuse/core'
 import type { Address } from 'viem'
 
 const props = defineProps<{
   address: Address
 }>()
 
-// const teamStore = useTeamStore()
-// const currencyStore = useCurrencyStore()
-
-const currency = useStorage('currency', {
-  code: 'USD',
-  name: 'US Dollar',
-  symbol: '$'
-})
 // Reactive state for balances: composable that fetches address balances
-const { balances, isLoading } = useContractBalance(props.address as Address)
+const { data: balance, isLoading } = useContractBalance(props.address as Address)
+
+const iconFor = (symbol: string) => {
+  if (symbol === 'USDC' || symbol === 'USDCe') return USDCIcon
+  if (symbol === 'POL') return MaticIcon
+  if (symbol === 'ETH') return EthereumIcon
+  return null
+}
+
+/**
+ * `price` and `balance` are flattened to plain numbers because the table sorts
+ * on the accessor value, and a `CurrencyPair` object would not compare. The
+ * display strings ride alongside as `*Label`.
+ */
+const rows = computed(() =>
+  (balance.value?.balances ?? []).map((entry, index) => ({
+    ...entry,
+    rank: index + 1,
+    icon: iconFor(entry.token.symbol),
+    price: entry.price.local.value,
+    priceLabel: entry.price.local.formatted,
+    balance: entry.value.local.value,
+    balanceLabel: entry.value.local.formatted
+  }))
+)
 </script>
--- a/app/src/components/forms/DepositBankForm.vue
+++ b/app/src/components/forms/DepositBankForm.vue
@@ -71,7 +71,7 @@ import { ref, computed } from 'vue'
 import TeamArchivedTooltip from '@/components/TeamArchivedTooltip.vue'
 import { z } from 'zod'
 import { parseEther, zeroAddress, type Address } from 'viem'
-import { useContractBalance } from '@/composables/useContractBalance'
+import { useContractBalance, contractBalanceKeys } from '@/composables/useContractBalance'
 import { useSafeSendTransaction } from '@/composables/transactions/useSafeSendTransaction'
 import { useERC20Approve } from '@/composables/erc20/writes'
 import { useErc20Allowance } from '@/composables/erc20/reads'
@@ -130,7 +130,8 @@ const userDataStore = useUserDataStore()
 const toast = useToast()
 
 // Reactive state for balances
-const { balances, isLoading } = useContractBalance(userDataStore.address as Address)
+const { data: balance, isLoading } = useContractBalance(userDataStore.address as Address)
+const balances = computed(() => balance.value?.balances ?? [])
 
 // Native token deposit using safe transaction handler
 const nativeDeposit = useSafeSendTransaction({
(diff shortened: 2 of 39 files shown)


hermannleboss:
Added the `large-pr-justified` label and a `## Large PR justification` section to the description.

Short version: this changes the public shape of a composable used by 16 components, so every split leaves the repo unable to type-check. A stacked base PR (composable + data structure) merged on its own would break `develop`, since all 16 consumers still destructure `{ balances, total }`. Extracting the new pure layer first would land 130 lines of production code with no caller. There is no behavior change to separate out, and everything is in `app/`.

The 843 test lines are fixture rewrites forced by the same shape change, and the diff is a net reduction overall — 294 insertions against 473 deletions. The four commits are individually reviewable and best read in order.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

hermannleboss:
Looked into it. The headline `9.75%` is the **app-e2e-tests** flag — e2e never exercises this code. The unit flag reports `92.70%` patch on the same diff, and both required checks (`codecov/patch`, `codecov/project`) pass.

Measured locally with istanbul, the two new files were already at **100% statements / functions / lines**, with exactly two uncovered *branches*. Both turned out to be unreachable defensive code, so I deleted them rather than writing tests for inputs the types forbid (`6358abc9`):

- `tokens = SUPPORTED_TOKENS` — defaulted a parameter the only caller always passes explicitly.
- `token.decimals ?? 18` — guarded a field `TokenConfig` declares as a required `number`.

`src/lib/balances/tokenBalances.ts` and `src/composables/useContractBalance.ts` are now at 100% on all four metrics.

The remaining entries in the report are one- and two-line template/expression changes in `.vue` files whose specs already cover the components' behaviour (`MainContractBalanceCell`, `CompanyOverview`), plus a comment-only invalidation change in `useCashOutAll` and a single field access in `expenseUtil`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2415, opened 2026-08-02
feat: ✨ add an "as of" date picker to the accounting Summary

### LINKED ISSUE #2103: feat: ✨ add an "as of" date picker to the accounting Summary
@
## Context

The Polymarket accounting **Summary** ([`AccountingSummary.vue`](dashboard/app/components/accounting/AccountingSummary.vue)) currently shows **all-time** metrics only (Total return, Profit/Loss, Net deposits, Portfolio value, Realized/Unrealized P&L, etc.). There is no way to see where the portfolio stood at a given date.

## Goal

Add a date picker to the Summary, reusing `AccountingDatePicker` in `mode="date"` ("as of" snapshot), exactly like the Balance Sheet. When the user picks a date, **all** Summary metrics must recompute as of that date.

## Acceptance criteria

- [ ] A date picker (mode `date`) is visible on the Summary, defaulting to "Today", with no regression to the current rendering.
- [ ] The existing `AccountingDatePicker` component is reused (no duplication).
- [ ] Changing the date recomputes the metrics bounded by that date.
- [ ] The selection persists across reload / tab switch (localStorage, dedicated key).

Georginio-prod:
The Summary was all-time only, with no way to see where the portfolio stood on a given date. It now carries the same date picker as the Balance Sheet (mode "date", default Today, selection persisted), and every figure is replayed from the dated feeds for the chosen date.

A past snapshot carries still-open bets at cost, exactly like the Balance Sheet does, because Polymarket publishes no historical market price. The figures that only exist for today (unrealized P&L, profile P&L, position basis drift) are shown as a dash with the reason rather than today's numbers under a past date. Picking Today returns the live summary untouched, so the default rendering is unchanged.

Card definitions moved out of the component into a util so it stays a description of the layout.

Closes #2103

### DIFF
--- a/dashboard/app/components/accounting/AccountingSummary.vue
+++ b/dashboard/app/components/accounting/AccountingSummary.vue
@@ -1,5 +1,22 @@
 <template>
   <div class="space-y-4">
+    <!-- Reporting date — "as of" snapshot, same picker as the Balance Sheet. -->
+    <div class="flex flex-col sm:flex-row sm:items-center justify-between gap-3">
+      <h3 class="font-semibold text-black dark:text-white">
+        Summary
+      </h3>
+      <AccountingDatePicker
+        v-model="asOf"
+        mode="date"
+        storage-key="dashboard-accounting-summary-asof"
+      />
+    </div>
+
+    <p v-if="!snapshot.isCurrent" class="text-xs text-muted">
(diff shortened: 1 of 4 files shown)


hermannleboss (commented):
LGTM

Georginio-prod (commented):
LGTM

PR #2420, opened 2026-08-04
feat: ✨ withdraw funds left on a team's archived contract generations

### LINKED ISSUE #2419: feat: ✨ withdraw funds left on a team's archived contract generations
Parent goal: #2263

## Problem

When a team redeploys its Officer, the previous generation stays on-chain and its contracts keep whatever they held. The redeploy modal already warns about it:

> Funds held in the **old** Bank, Expense Account and Cash Remuneration contracts are still on-chain but are no longer reachable through the new Officer's UI. Withdraw or migrate them first if you need them.

Today there is no "first" to do it in: nothing in the app can move those funds. The contract management page lists the archived generations and even shows their balances, but offers no action on them, so the money is stranded unless the owner drives the contracts by hand.

## Proposal

Add a **Withdraw funds** action on each archived Officer generation card in the contract management page. It runs the same three-step sequence the dashboard "Cash out all" already uses, but scoped to the archived generation and pointed at the team's current Bank:

1. old Cash Remuneration → sweeps into the Bank of its own generation
2. old Expense Account → sweeps into the Bank of its own generation
3. old Bank → forwards native + every held ERC-20 to the **current** Bank

The money therefore stays in the team treasury instead of landing in someone's wallet. Only the final hop is configurable — on-chain, `ownerWithdrawAllToBank` resolves the Bank through the Officer that deployed the contract, so the sweeps always land in their own generation's Bank.

## Version constraint

`ownerWithdrawAllToBank` landed in generation V1. Each generation is a distinct full redeployment behind its own frozen beacons, so V0 / V0.1 proxies can never gain it — steps 1 and 2 are impossible for them.

`Bank.transfer` / `transferToken` however have existed since V0, so **step 3 still works on every generation**. A V0 / V0.1 card therefore offers a Bank-only withdrawal rather than nothing at all, and states plainly that the Expense Account and Cash Remuneration stay behind. Emptying those two by hand (EIP-712 budget approval / signed wage claim) is out of scope here.

## Acceptance criteria

- [ ] Each archived generation card exposes a withdraw action that drains its Bank, Expense Account and Cash Remuneration into the team's current Bank
- [ ] A V0 / V0.1 generation still gets its Bank drained, and the review names the accounts left behind
- [ ] A review step shows the per-account balances and the destination before any signature
- [ ] Progress is reported per step, a failed step stops the sequence and can be retried without redoing the successful ones
- [ ] The action is unavailable — with a visible reason — when the generation cannot be resolved on-chain, when the connected user does not own the contracts, when the team is archived, or when nothing withdrawable is left
- [ ] Unit coverage for the version rule, the plan restriction, the address selection, the parameterized sequence, and the component's blocked states

hermannleboss:
Closes #2419

## What

Adds a **Withdraw funds** action to every archived Officer generation card on the contract management page. Redeploying the Officer leaves the previous generation on-chain holding whatever it held — the page already listed those balances but offered no way to reach them, so the money was stranded.

The action runs the three-step sequence the dashboard "Cash out all" already uses, scoped to the archived generation and pointed at the team's **current Bank**, so the funds stay in the treasury:

1. old Cash Remuneration → sweeps into the Bank of its own generation
2. old Expense Account → sweeps into the Bank of its own generation
3. old Bank → forwards native + every held ERC-20 to the current Bank

Only the last hop is configurable: on-chain, `ownerWithdrawAllToBank` resolves the Bank through the Officer that deployed the contract, so the sweeps always land in their own generation's Bank.

## How

Rather than fork the sequence, the existing pieces were made address-aware:

- `useTransfer` / `useTransferToken` / `useOwnerWithdrawAllToBank` / `useBankOwner` take an optional address override, defaulting to the team store exactly as before. Each hook still wraps one call.
- `useCashOutAll` takes optional `sources` + `to`, defaulting to the current generation and the connected wallet. The dashboard call site is untouched.
- `useOfficerBeaconFolderQuery` exposes `isPending` / `isError` alongside the folder, because an in-flight read and a failed one both surface as `undefined` and this decision must not treat either as "unknown generation".
- The progress stepper, previously restated in full in the dashboard flow, is now one presentational component keyed by a test prefix.

## Generations that predate owner withdrawal

`ownerWithdrawAllToBank` landed in generation **V1**, and each generation is a distinct full redeployment behind its own frozen beacons — a V0 / V0.1 proxy can never gain it.

`Bank.transfer` / `transferToken` have existed since **V0** though, so those generations are not blocked: their card offers a **Bank-only** withdrawal. The plan is restricted to the Bank step, the source addresses are withheld from the sequence so it never calls a function their proxies do not implement, and both the review and the completion notice name the accounts left behind — an empty Bank must never read as an empty generation. Emptying those two by hand (EIP-712 budget approval / signed wage claim) stays out of scope.

The version rule is a denylist (`V0`, `V0.1`), not an allowlist, so a future generation is supported by default — new generations only ever add to the ABI.

## Blocked states

Every disabled state names its reason in the tooltip: no Bank in the generation, no current Bank to receive, the generation already holds the current Bank, generation still resolving, generation unreadable, not the contract owner, team archived, nothing withdrawable left (worded differently for a Bank-only generation).

## Tests

`npm run lint`, `npm run build-only`, `npm run type-check`, `npm run format-check` and the full `npm run test:unit` (317 files, 2737 tests) all pass. New coverage: the version rule, the plan restriction for non-sweeping generations, the account lookup, the parameterized sequence (writes point at the supplied generation, balances read from the supplied Bank, funds forwarded to the supplied recipient, current generation untouched), and the component's blocked states, Bank-only path, review, progress, retry and completion paths.

## Not covered

Safe balances — the Safe is preserved across a redeploy and is not part of an archived generation.

### DIFF
--- a/app/src/components/CashOutStepList.vue
+++ b/app/src/components/CashOutStepList.vue
@@ -1,0 +1,59 @@
+<template>
+  <ul class="space-y-2">
+    <li
+      v-for="step in steps"
+      :key="step.key"
+      class="flex items-start gap-3 rounded-lg border border-gray-200 px-3 py-2 dark:border-gray-800"
+      :data-test="`${testPrefix}-step-${step.key}`"
+    >
+      <UIcon
+        :name="STEP_ICON[step.status]"
+        :class="['mt-0.5 size-5 shrink-0', STEP_ICON_CLASS[step.status]]"
+      />
+      <div class="min-w-0 flex-1">
+        <p class="font-medium">{{ step.label }}</p>
+        <p v-if="step.status === 'active' && step.detail" class="text-xs text-gray-500">
+          {{ step.detail }}
+        </p>
+        <p
+          v-if="step.status === 'failed' && step.error"
+          class="text-error text-xs"
+          :data-test="`${testPrefix}-error-${step.key}`"
+        >
+          {{ step.error }}
+        </p>
+      </div>
+    </li>
+  </ul>
+</template>
+
+<script setup lang="ts">
+import type { CashOutRunStep, CashOutStepStatus } from '@/composables/cashOut'
+
+/**
+ * Progress of a running cash-out sequence, one row per account.
+ *
+ * Purely presentational: every flow driving `useCashOutAll` renders the same
+ * stepper, so the markup and the status→icon mapping live here rather than
+ * being restated per call site. `testPrefix` keeps each flow's `data-test`
+ * hooks distinct.
+ */
+defineProps<{
+  steps: CashOutRunStep[]
+  testPrefix: string
+}>()
+
+const STEP_ICON: Record<CashOutStepStatus, string> = {
+  pending: 'i-heroicons-clock',
+  active: 'i-heroicons-arrow-path',
+  success: 'i-heroicons-check-circle',
+  failed: 'i-heroicons-x-circle'
+}
+
+const STEP_ICON_CLASS: Record<CashOutStepStatus, string> = {
+  pending: 'text-gray-400',
+  active: 'animate-spin text-warning',
+  success: 'text-success',
+  failed: 'text-error'
+}
+</script>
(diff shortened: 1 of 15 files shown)


hermannleboss (commented):
LGTM

PR #2431, opened 2026-08-05
docs: 📝 analyse the Community Credit (FixedReturn) user flow

### LINKED ISSUE #2430: docs: 📝 document the Community Credit (FixedReturn) user flow
## Context

Community Credit (the `FixedReturn` contract) is one of the larger features in the portal —
three views, 18 components, a dedicated store, its own read/write composables, a backend
metadata endpoint, and a repayment path that goes through `Bank`. None of it is documented.

The naming does not help: the contract is `FixedReturn`, the product is "Community Credit",
the UI calls a deployed instance a "Credit Account", and an on-chain `LendingOffer` is
rendered as a "round". Anyone picking the feature up has to rebuild that mapping from scratch.

There are also non-obvious behaviours that are easy to break on a refactor — the status
derivation uses the chain clock rather than the browser clock, `Refundable` does not mean a
refund is pending, `overdue` is cosmetic, and `FixedReturn` is a pass-through that only holds
lender funds while a round is still raising.

## Goal

Produce a single reference document covering the feature end to end: what the user does, what
the contract enforces, and where the two drift apart — with diagrams for the contract logic.

## Scope

- [ ] Naming map and role model (issuer vs lender, derived from on-chain ownership)
- [ ] Full state machine: the 4 on-chain `OfferState` values → the 7 UI `RoundStatus` values
- [ ] Custody model — where the tokens sit at each phase
- [ ] The four user flows: prerequisite gate, create a credit call, lend, deadline resolution,
      repayment
- [ ] Mermaid diagrams for the contract logic (lifecycle, custody, happy path, `lendFunds` guards)
- [ ] Findings: gaps between the UI flow and what the contract actually allows, ranked by
      consequence, each with a suggested direction
- [ ] Index entry in `docs/00_DOCUMENTATION_INDEX.md`

## Out of scope

Fixing the findings. The document records them and ranks them; each one that is worth acting on
gets its own issue.

hermannleboss:
# Description

## Intial Issue Description

Closes #2430 — Community Credit (the `FixedReturn` contract) is one of the larger features in
the portal and none of it is documented. Three views, 18 components, a dedicated store, its own
read/write composables, a backend metadata endpoint, and a repayment path that routes through
`Bank`. The naming does not help either: the contract is `FixedReturn`, the product is "Community
Credit", the UI calls a deployed instance a "Credit Account", and an on-chain `LendingOffer` is
rendered as a "round".

Adds `docs/features/community-credit/user-flow-analysis.md` — a single end-to-end reference for
the feature — plus an entry in `docs/00_DOCUMENTATION_INDEX.md`.

**Documentation only. No source file is touched.**

### What the document covers

- **Naming map** — the four names for the same concept, and the `offerId` ⇄ `round.id` mapping.
- **Role model** — issuer vs lender, derived from `FixedReturn.owner()` rather than a toggle.
- **State machine** — the 4 on-chain `OfferState` values mapped onto the 7 UI `RoundStatus`
  values, with the two subtleties that are easy to get wrong: `Refundable` does not mean a
  refund is pending (the principals are already back by the time a client can observe that
  state), and `overdue` is cosmetic (`repayLenders` never gates on `maturityDate`).
- **Custody model** — where the tokens physically sit at each phase. `FixedReturn` is a
  pass-through, not a vault: it holds lender funds only between the first deposit and the target
  being reached, which is exactly the window in which a refund is reachable.
- **The user flows** — prerequisite gate, create a credit call, lend, deadline resolution
  (refund vs accept partial funding), repayment — each with the guards the contract actually
  enforces alongside the ones the UI mirrors.
- **What is solid** — recorded deliberately so it does not get refactored away by accident:
  the chain-clock status derivation, per-offer token decimals, UI gates that mirror contract
  reverts rather than re-deriving them.

### Contract logic diagrams

Four mermaid diagrams in §10:

| # | Type | Shows |
|---|---|---|
| 10.1 | `stateDiagram-v2` | offer lifecycle — every transition with its function and modifier |
| 10.2 | `flowchart LR` | token custody — lender → FixedReturn → Bank → lenders |
| 10.3 | `sequenceDiagram` | happy path end to end: create / fund / repay |
| 10.4 | `flowchart TD` | the ordered `lendFunds` guards, through to the Bank sweep |

All four were checked against mermaid v11's own parser rather than eyeballed.

None fixed here — deliberately. §8 records **seven findings** where the UI flow and the contract
diverge, ranked by consequence, each with a suggested direction. The one worth a reviewer's
attention:

> **F1** — `NewView.vue:302-321` publishes in two steps: `createLendingOffer` on-chain, then a
> metadata `POST` for the title. The `offerId` is inferred by re-reading `getTotalOfferings()`
> after the write, which is only correct if no other create landed in between. And if the `POST`
> fails, the user is left on the wizard with a Publish button while the offer already exists
> on-chain — clicking again sends a *second* real round. This is the only path in the feature
> where ordinary user behaviour (a flaky network) can produce wrong on-chain state.
>
> The fix is cheap: `mutateAsync` already resolves to `{ hash, receipt }` and the contract emits
> `LendingOfferCreated(offerId, …)`, so decoding the event removes the race entirely.

The remaining six are UI/navigation debt (a design-prototype layout switcher shipped to
production as the entry point to the repayment flow, `variant` held as global state instead of a
route, funded/overdue rounds filed under "History", no lender-facing view, missing token-balance
invalidations, and a latent owner-mismatch on the Repay gate).

**Reviewer note:** the findings are the part worth disagreeing with. Each is stated with a file
and line so it can be checked directly, and F1 should get its own issue if you agree with the
read.

## Type of change

- [ ] Tests for the changes have been added (for bug fixes/features) — N/A, documentation only

## Large PR justification

615 lines, all of them one documentation file (`user-flow-analysis.md`) plus a single-line index
entry. **No source file is touched** — `git diff --stat` against `develop` is two files, both
under `docs/`.

Why it does not split usefully:

- It is one reference document, and the sections are load-bearing on each other. The findings in
  §8 cite the flows described in §7; the diagrams in §10 illustrate the state machine defined in
  §5; the custody model in §6 is what makes the refund reasoning in §7.3 correct. Half a
  reference document is not a smaller review — it is an incoherent one that has to be re-read
  once the other half lands.
- The reviewer-attention argument the budget is protecting does not transfer to prose. There is
  nothing to run, no regression surface, no coverage to assess, no interaction with existing
  behaviour. It reads linearly, and every claim carries a `file:line` so it can be checked
  independently rather than taken on trust.

Note for whoever maintains `dangerfile.js`: the budget counts hand-written prose as production
lines — `IGNORED_PATTERNS` excludes generated artifacts, lockfiles and snapshots, but has no
notion of documentation. A `docs/**` exclusion (or a separate, larger docs budget) would stop
documentation PRs from consuming the override. Deliberately not changed here — that is a CI
policy decision, not something to slip into a docs PR.

### DIFF
--- a/docs/00_DOCUMENTATION_INDEX.md
+++ b/docs/00_DOCUMENTATION_INDEX.md
@@ -81,6 +81,7 @@ Implement on-chain logic:
 - [Authentication](./auth/README.md) — SIWE, JWT, user management
 - [Statistics Feature](./features/stats/functional-specification.md) — Platform analytics implementation
 - [API Reference](./features/stats/stats-api.md) — Statistics API endpoints
+- [Community Credit — User Flow Analysis](./features/community-credit/user-flow-analysis.md) — FixedReturn lending rounds: state machine, custody model, flows, findings
 
 ### Developer Support
 
--- a/docs/features/community-credit/user-flow-analysis.md
+++ b/docs/features/community-credit/user-flow-analysis.md
@@ -1,0 +1,614 @@
+# Community Credit (FixedReturn) — User Flow Analysis
+
+**Date:** 2026-08-03 · **Scope:** `develop` · **Contract version:** `FixedReturn` v1.3.0
+
+A read of the whole feature end to end — Solidity, frontend, backend — describing what a
+user actually does, what the contract actually enforces, and where the two drift apart.
+
+---
+
+## 1. Naming
+
+There are three names for one thing, which is the first source of confusion:
+
+| Name                 | Where it appears                                              |
+| -------------------- | ------------------------------------------------------------- |
+| **FixedReturn**      | the contract, the ABI, every composable and query key         |
+| **Community Credit** | the product name — sidebar, routes, page titles               |
+| **Credit Account**   | the UI's word for _this team's deployed FixedReturn instance_ |
+
+And one more, internal to the mapping layer: an on-chain **lending offer** is rendered as a
+**round** in the UI (`offerId` ⇄ `round.id`, same number).
+
+Throughout this document: **issuer** = the FixedReturn owner (the team), **lender** = a member
+who funds a round.
+
+---
+
+## 2. What the feature is
+
+A team raises working capital from its own members. Each **round** has a funding target, a
+**flat** interest rate over the whole term (not annualised), a subscription deadline, and a
+maturity date. Members lend ERC-20 tokens before the deadline; if the target is reached the
+principal is swept into the team treasury; at maturity the team repays principal + interest,
+pushed back to every lender pro-rata.
+
+Two things follow from this design and explain most of the UI:
+
+- **Everything is push, never pull.** A lender never claims, never withdraws, never signs
+  anything after lending. Refunds and repayments are transactions the _issuer_ sends, which fan
+  out to every lender in the same call. So the lender-facing surface is deliberately thin.
+- **FixedReturn is a pass-through, not a vault.** It holds lender deposits only while a round
+  is still raising. The moment the target is hit, the whole principal leaves for the Bank. See
+  §6.
+
+---
+
+## 3. Architecture map
+
+| Layer          | Location                                                                                                                                                    | Responsibility                                       |
+| -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------- |
+| Contract       | [`contract/contracts/FixedReturn.sol`](../../../contract/contracts/FixedReturn.sol) (731 l.)                                                                | offers, deposits, refunds, repayment fan-out         |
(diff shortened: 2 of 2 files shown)


hermannleboss:
Danger size warning: added the `large-pr-justified` label and a `## Large PR justification` section — the 615 lines are one documentation file plus a one-line index entry, no source touched, and the sections cross-reference each other so splitting produces two incomplete halves rather than two smaller reviews. Danger now reports 0 warnings. Also flagged in the justification that `dangerfile.js` counts hand-written prose as production lines (`IGNORED_PATTERNS` has no `docs/**` notion) — left unchanged here, since that is a CI policy call.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2444, opened 2026-08-07
fix: 🐛 payroll claim list and wage-rate switch UI issues

### LINKED ISSUE #2445: fix: 🐛 payroll claim list and wage-rate switch UI issues
## Description

Several small but blocking UI problems in the Payroll area and in the member wage setup.

**1. Claim status doesn't refresh between the two payroll views**

When a weekly claim is signed (or submitted / withdrawn) in Company Payroll, the new status does not appear in My Payroll History — and the other way around — until the page is manually reloaded.

**2. Action (three-dots) menu on the claim list**

- Several menus could be open at the same time. Opening one should close the others.
- The rows inside an open menu were only clickable on the text, not across the whole row.
- On the last rows of the table, the menu was cut off at the bottom edge of the card.

**3. Wage rate switch turns itself off while typing**

In the member wage setup, when a rate is switched on and you type a value that starts with a zero (for example `0.1` or `0.01`), the switch turns itself off and the field becomes read-only, forcing you to re-enable it several times before you can finish entering the amount.

**4. Community Credit accounting is duplicated**

A loan is booked into an intermediate account and then transferred to the bank — two entries for one thing. It should be recorded once, directly on the bank.

## To Reproduce

1. Open Payroll → Company Payroll, sign a claim, then open My Payroll History without reloading → the status is stale.
2. On the claim list, open one three-dots menu, then another → both stay open. Try clicking the padded area of a menu row → nothing happens. Open the menu on a bottom row → it is clipped.
3. Open the member wage setup, switch a rate on, and type `0.1` → the switch turns off mid-entry.

## Expected behavior

- Claim status stays in sync across both payroll views without a manual reload.
- Only one action menu is open at a time, the whole menu row is clickable, and the menu is never cut off.
- The rate switch stays on while typing any value; it only changes when the user toggles it.

## Additional context

Frontend-only. All three are addressed in the linked PR.

Georginio-prod:
# Description

## Summary

- **Claim status now stays in sync across the two payroll views.** Signing, submitting or withdrawing a claim in Company Payroll is reflected in My Payroll History (and vice-versa) without a manual page reload. The shared list now refreshes when you navigate to the view that was invalidated.
- **Action (three-dots) menu on the claim list** — three separate issues fixed:
  - only one menu can be open at a time (opening one closes any other);
  - the whole menu row is clickable, not just the text;
  - the menu is rendered above the table so it is never clipped by the card, and it stays within the screen (bottom rows no longer cut off).
  - Also corrected a stale cache key so the Enable / Disable actions actually refresh the list.
- **Wage rate switch no longer turns itself off while typing.** Entering a value that starts with a zero (e.g. `0.1`, `0.01`) keeps the switch on; the switch only changes when the user toggles it.
- **Community Credit accounting simplified.** A loan is now recorded once, directly on the bank, instead of being booked into an intermediate account and then transferred out — no more duplicated entry.

The menu logic (shared open-state + positioning) was extracted into a small composable to keep the component within the line limit.

Fixes #2445

## Type of change

## Tests

Unit tests updated/added for the affected components (payroll dropdown, enable action, and both wage setup steps). `type-check`, `lint` and the unit test suite pass in `app/`.


Georginio-prod (commented):
LGTM

PR #2480, opened 2026-08-12
fix: 🐛 wage change mid-week splits the week and bypasses hour limits

### LINKED ISSUE #2479: fix: 🐛 wage change mid-week causes split WeeklyClaims and cap bypass
## Problem

When a team owner changes a member's hourly rate in the middle of a week, and
that member has already logged hours for that week, the week gets split in two.

Three things go wrong from there:

- **The hour limits restart from zero.** The hours already logged sit on one half
  of the week and the new ones on the other, so neither half sees the full total.
  A member with a 40h weekly cap can end up logging well past it.
- **The owner sees the same week twice** in the list of weeks to approve.
- **One week gets paid twice**, as two separate payments at two different rates.

The same split happens from the other direction. If a member forgets to log their
hours and catches up later, after their rate has changed, the catch-up is filed
against today's rate instead of the rate that applied when the work was done — and
it opens a second copy of that week rather than joining the one already there. If
that week had already been approved, the guard that should block further changes
does not fire, because it is looking at the wrong copy.

## What should happen

A rate change should take effect at the **start of the next week**, so a week is
never half on the old terms and half on the new. Until that date, the current
rate and hour limits keep applying — including to hours logged late for earlier
weeks.

And the rate used for any set of hours should be the one that applied **during the
week the work was done**, not the one in force on the day it was typed in.

## Acceptance criteria

- [ ] Changing an existing rate schedules it for the start of the next week; the
      current rate holds until then
- [ ] A member's very first rate still applies immediately, so a new joiner is not
      blocked until Monday
- [ ] Hours logged late for an earlier week are paid at that week's rate, use that
      week's limits, and join the week already on file instead of creating a second one
- [ ] A week that was already approved, withdrawn or disabled refuses further hours,
      including when they are logged late
- [ ] The owner can correct a pending change before it takes effect without pushing
      its date back, and can cancel it entirely
- [ ] Both the member table and the member's own view show the pending change and its
      date, and switch over on their own once it takes effect
- [ ] Existing rates keep behaving exactly as they do today

Georginio-prod:
# Description

Rate changes now take effect at the **start of the next week**, so a week is never half on the old terms and half on the new. A member's first rate still applies immediately. The rate and limits used for any hours are the ones in force **during the week the work was done**, not on the day it was typed in — so hours logged late land on the week already on file instead of opening a second copy, and approved weeks stay closed. A pending change can be corrected without pushing its date back, or cancelled; owner and member both see it coming, and the switch happens on its own.

Two problems surfaced along the way and are fixed here: the screen could show a different rate than the server actually charged, and the claim calendar was silently disabling valid dates.

Fixes #2479

## Type of change

### DIFF
--- a/app/src/components/RateDotList.vue
+++ b/app/src/components/RateDotList.vue
@@ -17,14 +17,14 @@
           'bg-purple-500': !['native', 'usdc', 'usdt'].includes(rate.type)
         }"
       />
-      {{ rate.type === 'native' ? NETWORK.currencySymbol : rate.type.toUpperCase() }}
+      {{ rateSymbol(rate.type) }}
       {{ formatRateAmount(rate.amount) }}
     </div>
   </div>
 </template>
 
 <script setup lang="ts">
-import { NETWORK } from '@/constant'
+import { rateSymbol } from '@/utils/wageUtil'
 
 interface RateItem {
   type: string
--- a/app/src/components/sections/CashRemunerationView/Form/__tests__/ClaimForm.spec.ts
+++ b/app/src/components/sections/CashRemunerationView/Form/__tests__/ClaimForm.spec.ts
@@ -203,10 +203,11 @@ describe('ClaimForm.vue', () => {
       .findComponent({ name: 'UCalendar' })
       .props('isDateDisabled') as DateDisabledFn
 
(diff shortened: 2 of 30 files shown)


Georginio-prod (commented):
LGTM

Georginio-prod:
image
image
image

PR #2489, opened 2026-08-15
fix: 🐛 hold a wage change back only when hours are already submitted

### LINKED ISSUE #2479: fix: 🐛 wage change mid-week causes split WeeklyClaims and cap bypass

Georginio-prod:
# Description

- **Submitted hours hold a week to its wage** — nothing else does. A week with no hours in it, even
  one carrying weekly goals, takes the change whole from its own Monday, days already worked
  included.
- **A week is found by member and week, never by wage**, so it always has one row and cannot split.
  When the first hours land on a goals-only week, the row moves onto the wage that prices them.
- **Nothing is displayed when a change just takes effect.** The set-wage window warns only when the
  change has to wait.
- **A wait ends when its reason does.** The effective date is recomputed on each save, so a change
  queued while the week held hours stops waiting once it no longer does. It can only move forward,
  never back a week.

Left open, and noted in the feature notes: no database constraint forbids two rows for one
member-week — the lookup does. Making it structural needs a migration that first cleans up rows the
original bug may have left behind.

Fixed  #2479
## Type of change

## Validation

`backend`: lint clean, 778 unit tests. `app`: lint clean, 202 tests across the affected views. Docs:
markdown lint and format clean.

### DIFF
--- a/app/src/components/sections/DashboardView/SetMemberWageModal.vue
+++ b/app/src/components/sections/DashboardView/SetMemberWageModal.vue
@@ -27,13 +27,18 @@
 
       <template #body>
         <div class="mt-1 space-y-4">
+          <!--
+            Shown only when the change has to wait. A change that simply takes
+            effect is not news: the new wage is what the member's week will be
+            priced at, and saying so would be noise on the common path.
+          -->
           <UAlert
-            v-if="props.wage"
+            v-if="showEffectiveDateNotice"
             icon="i-heroicons-information-circle"
-            :color="changeTiming.immediate ? 'warning' : 'info'"
+            color="info"
             variant="soft"
             data-test="wage-effective-date-notice"
-            :title="effectiveDateTitle"
+            :title="`This change takes effect on ${changeTiming.label}.`"
             :description="effectiveDateDescription"
           />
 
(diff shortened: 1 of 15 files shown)


Georginio-prod (commented):
LGTM

